PR #73, opened 2019-06-24
PMM-4230 Adding PS Specific Bats tests

puneet0191:


pmm-tests/ps-specific-tests-2-x.bats
```diff
@@ -0,0 +1,75 @@
...
+    [ "$status" -eq 1 ]
+    [ "${lines[0]}" = "usage: pmm-admin [<flags>] <command> [<args> ...]" ]
+}
+
+
+@test "run pmm-admin add mysql based on running intsances" {
+	  COUNTER=0
+		for i in $(pmm-admin list | grep "MySQL" | sed 's|.*(||;s|)||') ; do
+      let COUNTER=COUNTER+1
+			run pmm-admin add mysql --username=${MYSQL_USER} localhost:3306 mysql_$COUNTER
```
Nailya: @puneet0191  would it be better to use some $PORT variable , so we can use this test for MySQL instance is installed on specified port.

pmm-tests/ps-specific-tests-2-x.bats
```diff
@@ -0,0 +1,75 @@
...
+	    	echo "${lines[0]}" | grep "MySQL Service added."
+		done
+}
+
+
+@test "run pmm-admin add mysql again based on running instances" {
+	COUNTER=0
+	for i in $(pmm-admin list | grep "MySQL" | grep "mysql_" | sed 's|.*(||;s|)||') ; do
+		let COUNTER=COUNTER+1
+		run pmm-admin add mysql:metrics --username=${MYSQL_USER} mysql__$COUNTER
```
Nailya: the same problem with port here

PR #79, opened 2019-07-01
PMM-4230 Adding MongoDB Bats tests

puneet0191:


pmm-tests/pmm-2-0-bats-tests/mongodb-tests.bats
```diff
@@ -1,107 +1,77 @@
...
+		run pmm-admin add mongodb --use-profiler ${MONGO_IP_PORT} mongo_inst_${COUNTER}
+	  [ "$status" -eq 0 ]
+	  echo "${lines[0]}" | grep "MongoDB Service added"
+  done
+}
+
+@test "run pmm-admin add mongodb again based on running instances" {
+	COUNTER=0
+	IFS=$'\n'
+	for i in $(pmm-admin list | grep "MongoDB" | grep "mongo_inst_" | awk -F" " '{print $3}') ; do
```
Nailya: @puneet0191 I guess it should be changed to " pmm-admin list | grep "MongoDB" | grep "mongo" | awk -F" " '{print $3}'"

puneet0191: nope this works fine because I am adding instances with that name before this, it is just to reconfirm addition with same name doesn't work

Nailya: oh, sorry, I missed it :)

PR #85, opened 2019-07-09
Pmm 4304 wipe clients for pmm2

Nailya:
added wipe-clients for PMM2 and removed redundant client's configuration

### DIFF
--- a/pmm-tests/pmm-framework.sh
+++ b/pmm-tests/pmm-framework.sh
@@ -1659,22 +1659,42 @@ clean_clients(){
    exit 1
   fi
   #Shutdown all mysql client instances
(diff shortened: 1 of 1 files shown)


pmm-tests/pmm-framework.sh
```diff
@@ -1841,9 +1860,9 @@ if [ ! -z $setup ]; then
   setup
 fi
 
-if [ ! -z $PMM2 ]; then
-  configure_client;
-fi
+#if [ ! -z $PMM2 ]; then
```
puneet0191: Why comment on this?

Nailya: I removed it, because of client's configuration is called in installation of pmm2-client (see 700 line) and I think we don't need to call it each time. I checked installation of PMM server and adding clients and everything worked ok without this part

PR #94, opened 2019-07-17
PMM-4309 Adding support for MS in testsuite, using updated wipe clients

puneet0191:


pmm-tests/pmm-framework.sh
```diff
@@ -1408,9 +1408,9 @@ add_clients(){
           sudo sed -i "s/3306/${PXC_PORT}/" /etc/proxysql-admin.cnf
           sudo proxysql-admin -e > $WORKDIR/logs/proxysql-admin.log
           if [ $disable_ssl -eq 1 ]; then
```
Nailya: do we still need this? As I know we didn't  add --disable-ssl  option to PMM2 client

puneet0191: yes, It is not being used at the moment, I just kept it in case we add it in the future.

PR #125, opened 2019-08-22
PMM-4525 Adding Support for MongoDB official

puneet0191:


pmm-tests/pmm-framework.sh
```diff
@@ -49,6 +49,7 @@ usage () {
   echo " --pmm2-server-ip               Pass Address for PMM2-Server"
   echo " --package-name                 Name of the Server package installed"
   echo " --ps-version                   Pass Percona Server version info"
+  echo " --modb-version                 Pass MongoDB version info, from MongoDB!!"
```
Nailya: @puneet0191 why not to use --mo-version parameter?

puneet0191: --mo is reserved for Percona's MongoDb distribution.

PR #221, opened 2020-02-28
PMM-5476 unskipped --pprof tests, fixed failing tests

yurkovychv:


pmm-tests/pmm-2-0-bats-tests/generic-tests.bats
```diff
@@ -195,8 +185,6 @@ run pmm-admin summary --skip-server
 echo "$output"
     [ "$status" -eq 0 ]
     echo "${lines[-1]}" | grep ".zip created."
-    checkZipFileContents
```
BupycHuk: let's keep this test.

yurkovychv: @BupycHuk  this test exists, but we decided to remove files count check. or you think we should check it too?

BupycHuk: We don't have other test with the same amount of files

yurkovychv: we do actually, at line 174

BupycHuk: you are right, my bad

PR #230, opened 2020-03-03
PMM 5434 added autodelay and allure plugins

yurkovychv:


local.codecept.json
```diff
@@ -66,6 +66,12 @@
       "browsers": ["chrome"]
     }
   },
+  "plugins": {
+    "allure": {},
```
puneet0191: This would require us to also install allure  on builds, why do we need allure with autodelay plugin?

yurkovychv: removed it

PR #235, opened 2020-03-09
PMM-5429 Cleanup and Updating Readme with Selenoid Setup 

puneet0191:


README.md
```diff
@@ -2,7 +2,18 @@
 Automated tests for Percona Monitoring and Management
 GUI tests are created for testing frontend of PMM. They include tests for Query Analytics and for Grafana dashboards
 
-## Installation (UI tests version 2.0)
+## Using Selenoid for running tests in Local
+1. Install Nodejs and atleast npm 8.x on your system
```
atymchuk: Nodejs should be `Node.js`
typo `at least`

README.md
```diff
@@ -2,7 +2,18 @@
...
+2. Selenoid and Selenoid UI uses port 4444, 8080, make sure they are not being used, other change docker-compose.yml file.
```
atymchuk: `Selenoid and Selenoid UI use port 4444 and 8080 respectively, make sure they are not being used, otherwise update docker-compose.yml file`

README.md
```diff
@@ -2,7 +2,18 @@
...
+3. run npm install in project root.
+4. run prepare_ui_test.sh script in the root directory.
+`bash -x ./prepare_ui_test.sh`
+5. This should start running UI tests in 4 parallel browser session inside chrome container with selenoid
```
atymchuk: typos `sessions`, `inside a Chrome container with Selenoid`

README.md
```diff
@@ -2,7 +2,18 @@
...
+6. Check live execution by going to localhost:8080 port.
```
atymchuk: Does it mean `...by launching http://localhost:8080 in your browser` or something else?

README.md
```diff
@@ -2,7 +2,18 @@
...
+
+## Looking for a detailed and deeper way to run UI tests, checkout next sections.
```
atymchuk: I'd suggest `If you'd like to have more control over the UI test framework parameters, please check out next sections`

prepare_ui_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+### Download Docker Images
+echo "you must have permission to run docker."
```
atymchuk: I'd suggest to either have a script check wether the user has a permission to run docker, or to echo it as a `Warning: ...`

prepare_ui_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+docker pull selenoid/vnc_chrome:80.0
+docker pull selenoid/video-recorder:latest-release
+docker pull aerokube/selenoid-ui
+
+###create env file
+echo PWD=${PWD} > .env
+
+###
+docker-compose up -d
+sleep 20
```
atymchuk: May not be enough on slower machines. Could we use a health check maybe?
Pls see if you can adapt this (credit @AlekSi):  `docker run --health-cmd='curl -f http://127.0.0.1/ping || exit 1' --health-interval=3s --health-interval=2s --health-start-period=10s --health-retries=3`

README.md
```diff
@@ -2,7 +2,18 @@
...
+
+### Installation (UI tests version 2.0)
```
atymchuk: Same as above for Nodejs et al

prepare_ui_tests.sh
AlekSi: It can be added directly to docker-compose.yml: https://docs.docker.com/compose/compose-file/#healthcheck

prepare_ui_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+### Download Docker Images
+echo "you must have permission to run docker."
+docker pull selenoid/vnc_chrome:80.0
+docker pull selenoid/video-recorder:latest-release
+docker pull aerokube/selenoid-ui
```
AlekSi: It can be `docker-compose pull` – it can pull several images in parallel.

prepare_ui_tests.sh
AlekSi: But then there should be `until healthy` loop anyway.

puneet0191: it is a good suggestion but not really wanting to do it right in this revision, thank you.

prepare_ui_tests.sh
puneet0191: vnc_chrome and video-recorder is not really part of compose, it is being used by selenoid container, so docker-compose pull will only pull selenoid-ui and selenoid, so this will probably be needed.

README.md
```diff
@@ -2,8 +2,20 @@
 Automated tests for Percona Monitoring and Management
 GUI tests are created for testing frontend of PMM. They include tests for Query Analytics and for Grafana dashboards
 
-## Installation (UI tests version 2.0)
-1. Install Nodejs and atleast npm 8.x on your system
+## Using Selenoid for running tests in Local
+1. Install Node.js and atleast npm 8.x on your system
```
atymchuk: typo `at least`

prepare_ui_tests.sh
atymchuk: Ok, will add later.

PR #236, opened 2020-03-09
Pmm 5434 Update QAN UI tests to check Information for Tables in PG

yurkovychv:
added tests for Postgres in QAN, tests for Updating/Clearing Alertmanager rules, updated change Metrics resolution test (removed Change Data Retention step)

### DIFF
--- a/tests/pages/pmmSettingsPage.js
+++ b/tests/pages/pmmSettingsPage.js
@@ -23,7 +23,19 @@ module.exports = {
             "      annotations:\n" +
             "        summary: \"Instance {{ $labels.instance }} down\"\n" +
             "        description: \"{{ $labels.instance }} of job {{ $labels.job }} has been down for more than 20 seconds.\"",
-        ruleName: "AutoTestAlerts"
+        editRule: "groups:\n" +
+            "  - name: AutoTestAlertsEdited\n" +
+            "    rules:\n" +
+            "    - alert: InstanceDown\n" +
+            "      expr: up == 0\n" +
+            "      for: 60s\n" +
+            "      labels:\n" +
+            "        severity: critical\n" +
+            "      annotations:\n" +
+            "        summary: \"Instance {{ $labels.instance }} down\"\n" +
+            "        description: \"{{ $labels.instance }} of job {{ $labels.job }} has been down for more than 20 seconds.\"",
+        ruleName: "AutoTestAlerts",
+        editRuleName: "AutoTestAlertsEdited"
     },
     messages:{
         successPopUpMessage: "Settings updated",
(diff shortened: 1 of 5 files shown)


tests/pages/pmmSettingsPage.js
```diff
@@ -191,7 +203,7 @@ module.exports = {
 
     customClearField(field) {
         I.appendField(field, '');
-        I.pressKey(['Shift', 'Home']);
+        I.pressKey(["Control", "a"]);
```
puneet0191: why

tests/verifyQAN_test.js
```diff
@@ -46,4 +46,52 @@ Scenario('Open the QAN Dashboard and verify pagination', async (I, adminPage, qa
...
+    qanPage._selectDetails(2);
+    qanPage.selectSectionInDetails(qanPage.fields.explainTabInDetails);
+    await qanPage.verifyExplainDetailsSectionDataExists();
+});
+
+Scenario('Open the QAN Dashboard and verify Explain tab in Query Details for Environment=pgsql-dev filter', async (I, adminPage, qanPage) => {
+    I.amOnPage(qanPage.url);
+    await I.waitForElement(qanPage.fields.iframe, 60);
+    adminPage.applyTimer("5m");
+    await I.switchTo(qanPage.fields.iframe); // switch to first iframe
```
puneet0191: when you switch to Iframe, when do you switch back?

tests/pages/qanPage.js
```diff
@@ -125,8 +133,37 @@ module.exports = {
...
         I.click("//table/tr["+ (row + 1) + "]//td[2]");
+        this.waitForDetailsSection();
+    },
+
+    selectSectionInDetails(section) {
+        I.waitForElement(section, 30);
+        I.click(section);
+    },
+
+    waitForTableTabContentsLoaded() {
```
puneet0191: The same feedback, I understand you want to increase readability, but having `waitForTableTabContentsLoaded` and `waitForExplainTabContentsLoaded` and then may be more in future, could you think of making a more generic function, wait for details and pass parameters? `waitForTabContentsLoaded` if it is not feasible, not a problem

tests/pages/pmmSettingsPage.js
yurkovychv: thats for clearing field. clearField method does not work correctly with React

tests/verifyQAN_test.js
yurkovychv: i do not need it

tests/pages/qanPage.js
yurkovychv: yes, i'll fix it as you say

PR #287, opened 2020-05-08
Default PMM-Server version set to dev-latest, if not provided explicitly

meet59patel:
When **--pmm-server-version** option is not explicitly used to provide version, and also if the **--dev** option is provided then PMM-Server version will be set to **dev-latest** 

Note: dev-latest Docker tag works only for perconalab/pmm-server repository on Dockerhub.
@puneet0191 Please review.

### DIFF
--- a/pmm-tests/pmm-framework.sh
+++ b/pmm-tests/pmm-framework.sh
@@ -303,6 +303,10 @@ do
     --dev )
     shift
     dev=1
+    if [[ -z $pmm_server_version ]];then
+    	pmm_server_version="dev-latest"
+    	PMM_VERSION=$pmm_server_version
+    fi
     ;;
     --with-proxysql )
     shift


pmm-tests/pmm-framework.sh
```diff
@@ -303,6 +303,10 @@ do
     --dev )
     shift
     dev=1
+    if [[ -z $pmm_server_version ]];then
```
puneet0191, line 306: have you checked what happens between line 140 to 150 in this script?

puneet0191:
Good Start @meet59patel  I am really glad!, keep it up, I left a minor comment

pmm-tests/pmm-framework.sh
meet59patel: Yes I have. I think in the transient period where all the command line arguments are being looped through, order in which `--dev `and `--pmm-server-version` is provided would matter. But our main execution starts only after the loop has checked all the options provided so ordering wouldn't hurt at the end.

meet59patel: Actually I first placed my edit at the lines you mentioned, but then I realized that `dev-latest` works only with `perconalab` dockerhub repo, and not `percona` repo. So I decided to set the default value at `--dev` option.

meet59patel:
@puneet0191 Closing this PR as discussed previously. 
I'll implement suggested changes (moving the added lines to a better location). 
Also, contributions from me will be pushed to `GSOC-2020` branch henceforth.

PR #296, opened 2020-05-20
Remove UI tests

puneet0191:


AlekSi:
But shouldn't we create a FB? ;)

PR #368, opened 2020-12-18
External service option

puneet0191:


pmm-tests/pmm-framework.sh
```diff
@@ -2425,6 +2430,19 @@ setup_pmm2_client_docker_image () {
   sleep 5
 }
 
+setup_external_service () {
+  wget https://github.com/oliver006/redis_exporter/releases/download/v1.14.0/redis_exporter-v1.14.0.linux-386.tar.gz
+  tar -xvf redis_exporter-v1.14.0.linux-386.tar.gz
+  mv redis_* redis_exporter
```
Nailya, line 2436: probably redis_exporter directory should be created first

Nailya: or change to :
`mv redis_exporter-v1.14.0.linux-386 redis_exporter`

Nailya (changes requested):

pmm-tests/pmm-framework.sh
puneet0191: Nailya it will be done, when we extract the tar, and I M already renaming with wildcard

Nailya (approved):

PR #396, opened 2021-05-06
Add script for backup toolkit instalation

VoK0:
This will add script which will be responsible for installing compatible version of xtrabackup based on mysql version which was previously defined by aws-staging start job.

### DIFF
--- a/install_backup_toolkit.sh
+++ b/install_backup_toolkit.sh
@@ -5,6 +5,6 @@ if [[ \$msql_version == '5.7' ]]; then
 elif [[ \$msql_version == '8.0' ]]; then
     sudo yum install percona-xtrabackup-80  qpress -y
 else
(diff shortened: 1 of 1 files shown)


install_backup_toolkit.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+msql_version=$1
```
puneet0191, line 2: Could you also do a sanity check if only 1 parameter is passed, and handle with error message

puneet0191 (changes requested):

puneet0191:
PR Title add ticket number, make sure commit messages have ticket number, it help us track on Jira.

install_backup_toolkit.sh
VoK0: done please check latest version :)

PR #400, opened 2021-05-19
delete operators install

beatahandzelova:


puneet0191 (approved):

puneet0191:
@beatahandzelova I think we can keep this script for use in the future for trying out some other versions, call for the script is removed from the pipeline is removed, hence closing this one. 

thanks a lot.

PR #496, opened 2022-07-20
PMM-9951 official Postgres and PGSS Support

puneet0191:


pmm-tests/pgsql_pgss_setup.yml
```diff
@@ -0,0 +1,81 @@
+---
+# This playbook does following:
+#   enables Percona testing repository
```
beatahandzelova, line 3: Can you please add correct description?

puneet0191: Done

beatahandzelova (approved):

PR #506, opened 2022-09-05
Add part when pmm-agent run as root

VoK0:
* fix pmm-agent permission issues for backup managment

### DIFF
--- a/pmm-tests/backup/ps_80_bm.yml
+++ b/pmm-tests/backup/ps_80_bm.yml
@@ -206,4 +206,3 @@
     become: yes
     become_user: root
     when: is_aws_environment
-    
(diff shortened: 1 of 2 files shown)


pmm-tests/backup/ps_80_bm.yml
```diff
@@ -159,3 +159,34 @@
...
+    become: yes
+    become_user: root
+
+  - name: run pmm-agent as root user
+    ignore_errors: yes
+    shell: pmm-agent --config-file=/home/ec2-user/workspace/aws-staging-start/pmm2-client/config/pmm-agent.yaml >> pmm-agent_log.txt
+    async: 2592000
+    poll: 0
+    become: yes
+    become_user: root
```
puneet0191, line 192: add empty last line

puneet0191: @VoK0 my only concern with this change is, it makes it difficult for anyone to setup locally, those paths are meant to work only on staging, can we try to find an alternative way?

puneet0191 (approved):
I have just shared my concern about the setup working only on aws based setups

puneet0191 (approved):

PR #519, opened 2022-10-18
PMM-10779 Add vacuum setup

peterSirotnak:
https://pmm.cd.percona.com/job/pmm2-ui-tests/23565/


pmm-tests/pgsql-vacuum.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+export DOCKER_CONTAINER_NAME=pgsql_vacuum_db
+echo "Setting up Postgres for vacuum monitoring"
+docker stop pgsql_vacuum_db || true
+docker rm pgsql_vacuum_db || true
+docker run --name pgsql_vacuum_db -p 7432:5432 -e POSTGRES_PASSWORD=YIn7620U1SUc -d postgres:14.5 \
```
puneet0191, line 7: all looks good to me, just I'd recommend the feature of passing version to this, we could setup vacuum for any DB version we support, we already have two variables set via pmm-framework --pdpgsql-version and --pgsql-version, so if user uses --setup-pgsql-vacuum --pgsql-version=1

puneet0191 (approved):
I think this is good, just one comment about allowing user passed version so I could setup vacuum with any supported version

pmm-tests/pgsql-vacuum.sh
peterSirotnak: Added story to add that functionality. [PMM-10912](https://jira.percona.com/browse/PMM-10912``)

PR #910, opened 2026-04-02
MCP use and Rules

kiranvuyurru:


.agents/workflows/pmmApi.json
```diff
@@ -0,0 +1,690 @@
+{
+  "base": "/v1",
+  "endpoints": [
```
yurkovychv, line 3: Is there a possibility to dynamically download the latest api?

kiranvuyurru: i can write a script to automatically update this file

yurkovychv: Would be great

yurkovychv: lets do it in a separate PR

yurkovychv (approved):

PR #911, opened 2026-04-06
PMM-14957 variable persistence

kiranvuyurru:


e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -16,7 +16,7 @@ export default class Dashboards extends BasePage {
   readonly valkey: ValkeyDashboardsType = ValkeyDashboards;
   builders = {
     panelByName: (panelName: string) =>
-      this.grafanaIframe().locator(`//section[contains(@data-testid, "${panelName}")]`),
+      this.grafanaIframe().getByTestId(`data-testid Panel header ${panelName}`),
```
yurkovychv: not sure if this is a good idea, there might be some panels that have service name in the panel name. So it is better to keep "old way"

yurkovychv (approved):

PR #951, opened 2026-05-05
PMM-15041: Add MongoDB in-memory engine option to PSMDB QA setup

travagliad:
## Summary

Adds optional MongoDB storage engine selection for the existing PSMDB QA flow so QA environments can run Percona Server for MongoDB with the in-memory storage engine and expose related PMM metrics.

## Changes

- Add `STORAGE_ENGINE` default (`wiredTiger`) to `PSMDB` framework options.
- Pass `MONGO_STORAGE_ENGINE` from `pmm-framework.py` into the `pmm_psmdb-pbm_setup` scripts.
- Parameterize `docker-compose-rs.yaml` mongod config bind mount via `MONGOD_RS_CONFIG_DIR` (defaults keep current WiredTiger config).
- Extend `start-rs-only.sh` to select WiredTiger vs inMemory and mount an in-memory `mongod.conf`.
- Add `conf/mongod-rs-inmemory/mongod.conf` with fixed `inMemorySizeGB` (not parameterized).

## Usage

```
python pmm-framework.py --database PSMDB=8.0,SETUP_TYPE=pss,STORAGE_ENGINE=inMemory
```

## Testing

- `python -m py_compile` on modified Python files
- `bash -n` on `start-rs-only.sh`
- `docker compose ... config` validation for `docker-compose-rs.yaml`

### DIFF
--- a/qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
+++ b/qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
@@ -3,17 +3,38 @@ set -e
 
 profile=${COMPOSE_PROFILES:-classic}
 mongo_setup_type=${MONGO_SETUP_TYPE:-pss}
+mongo_setup_type=${mongo_setup_type,,}
+mongo_storage_engine=${MONGO_STORAGE_ENGINE:-wiredTiger}
+mongo_storage_engine=${mongo_storage_engine,,}
 ol_version=${OL_VERSION:-9}
+
+
+export COMPOSE_PROFILES=${profile}
+export MONGO_SETUP_TYPE=${mongo_setup_type}
+export OL_VERSION=${ol_version}
+
+case "$mongo_storage_engine" in
+  wiredtiger)
+    ;;
+  inmemory)
+
+    generated_config_dir="/tmp/pmm-qa-mongod-rs-inmemory"
+    rm -rf "$generated_config_dir"
+    mkdir -p "$generated_config_dir"
+    cp ./conf/mongod-rs-inmemory/mongod.conf "$generated_config_dir/mongod.conf"
+    export MONGOD_RS_CONFIG_DIR="$generated_config_dir"
+    ;;
+  *)
+    echo "Unsupported MongoDB storage engine: $mongo_storage_engine"
+    exit 1
+    ;;
+esac
 
 docker network create qa-integration || true
 docker network create pmm-qa || true
 docker network create pmm-ui-tests_pmm-network || true
 docker network create pmm2-upgrade-tests_pmm-network || true
 docker network create pmm2-ui-tests_pmm-network || true
-
-export COMPOSE_PROFILES=${profile}
-export MONGO_SETUP_TYPE=${mongo_setup_type}
-export OL_VERSION=${ol_version}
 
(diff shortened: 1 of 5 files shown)


qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
```diff
@@ -3,18 +3,39 @@ set -e
 
 profile=${COMPOSE_PROFILES:-classic}
 mongo_setup_type=${MONGO_SETUP_TYPE:-pss}
+mongo_setup_type=${mongo_setup_type,,}
```
yurkovychv, line 6: Why this is needed?

qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
```diff
@@ -3,18 +3,39 @@ set -e
...
+mongo_storage_engine=${MONGO_STORAGE_ENGINE:-wiredTiger}
+mongo_storage_engine=${mongo_storage_engine,,}
 ol_version=${OL_VERSION:-9}
 
+
+export COMPOSE_PROFILES=${profile}
+export MONGO_SETUP_TYPE=${mongo_setup_type}
+export OL_VERSION=${ol_version}
+
+case "$mongo_storage_engine" in
```
yurkovychv: I think this is not needed. 
This makes the argument to be madatory. Pls fallback to wiredtiger as a default
The code should be much simpler in that case

yurkovychv (changes requested):

qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
travagliad: This is to normalize the string to lowercase, I can remove that

qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
travagliad: The argument isn't mandatory because I'm assigning wiredTiger as default, so if one don't fill anything, it still spin up the instance with the normal setup, only if you change it to inMemory it will work differently, but switch-case structure was too complex so I changed it to if-else

```bash
if [ "$mongo_storage_engine" = "inmemory" ]; then

    generated_config_dir="/tmp/pmm-qa-mongod-rs-inmemory"
    rm -rf "$generated_config_dir"
    mkdir -p "$generated_config_dir"
    cp ./conf/mongod-rs-inmemory/mongod.conf "$generated_config_dir/mongod.conf"
    export MONGOD_RS_CONFIG_DIR="$generated_config_dir"
else
    mongo_storage_engine="wiredTiger"
fi
```

qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
travagliad: This is important because it will work for "inMemory" which is the common name

qa-integration/pmm_psmdb-pbm_setup/start-rs-only.sh
travagliad: So to be clear, if you start `--database psmdb,SETUP_TYPE=pss`, it will work the exact same way as before

yurkovychv (approved):

PR #956, opened 2026-05-07
PMM-15033 add gh related scripts for nightly tests

yurkovychv:


codeceptjs-e2e/pr.codecept.js
```diff
@@ -18,6 +18,7 @@ exports.config = {
       restart: true,
       show: false,
       trace: true,
+      retry: 2,
```
travagliad, line 21: We added retry 2 as a default, but in many tests we're also enforcing retry(2), is it needed to keep those?

yurkovychv: updated, thanks

travagliad (approved):

PR #963, opened 2026-05-12
Add AGENTS.md for AI agents

ondrejpatocka:
## Summary
- Adds a root `AGENTS.md` - a single, authoritative entry point for AI agents working in `pmm-qa`.
- Complements (does not duplicate) the existing LLM workflow rules under [`.agents/workflows/`](.agents/workflows/).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,15 @@ All CI runs are GitHub Actions workflows under [.github/workflows/](.github/work
 - `fb-*.yml` — **feature-build** wrappers invoking a runner against a PR build.
 - `*-matrix*.yml` — matrix wrappers fanning a runner across versions/OS/arch.
 - `helm-tests.yml` — the only k8s entry point.
-- `rc-testing-suite.yml` — release-candidate orchestrator.
+- `rc-testing-suite.yml` — GitHub Actions portion of RC testing (see [External RC orchestration](#external-rc-orchestration) below).
 - `pmm-version-getter.yml` — reusable version-discovery helper.
 
 To find the entry workflow for a suite, search `runner-<suite>*.yml` in [.github/workflows/](.github/workflows/).
+
+### External RC orchestration
+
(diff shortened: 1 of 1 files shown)


yurkovychv (approved):

AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+This file is read by every AI agent at session start. **You are responsible for keeping it accurate.** After completing work, check whether any of these apply:
+
+- Added, removed, or renamed a top-level directory or test suite
+
+If any apply, update the relevant section of this file. Do **not** update for routine test additions or bug fixes that don't alter repo structure or conventions.
+
+## How This Documentation Is Organized
+
+This file is the **single authoritative entry point** for AI agents working with pmm-qa. It owns the product overview, repository map, and a consolidated per-suite quick-reference. Detailed suite docs live next to the suite's code (`README.md` / config file) — links are in the [Repository Map](#repository-map).
+
```
travagliad, line 14: Do this need to live in the agents file? AI will read this but this is not meant for AI but for humans instead. Can it be a part of our readme.md? A new section over AI contributing like you are instructing here? For me it makes sense to reduce information that goes to AI that is not needed

AGENTS.md
```diff
@@ -0,0 +1,87 @@
...
+    ghaPkg --> pkg
+    ghaHelm --> helm
+    Suites --> DUT["PMM Server + PMM Client (DUT)"]
+```
+
+`pmm-framework.py` is the common provisioning step for most CI jobs: it stands up PMM Client containers and monitored DBs on a shared Docker network named `pmm-qa`, then the respective UI / CLI suite runs against that environment.
+
+## CI / Pipelines
+
+All CI runs are GitHub Actions workflows under [.github/workflows/](.github/workflows/). Naming convention:
```
travagliad, line 78: Some CI runs are Github Actions and a lot others are Jenkins Pipelines, can we add the mention here?

You can feed AI with this file so it can automatically create info related to these other pipelines I'm mentioning, it's all grouped there:

https://github.com/Percona-Lab/jenkins-pipelines/blob/master/pmm/v3/pmm3-rc-testing.groovy

AGENTS.md
ondrejpatocka: Hi @travagliad ,
**Maintaining This Document**
- is an attempt to enforce AI agents to think of updating also Agents.md when proposing significant changes to repo, so would suggest to keep (same practice is applied in pmm/Agents.md file).

**How This Documentation Is Organized**
- I perceive this one as rather nice to have for faster agent alignment and orientation, can be removed

travagliad: In this case this makes total sense, thought it was for humans :D

travagliad (approved):

PR #989, opened 2026-05-29
PMM-7 fix @317 and @2205

kiranvuyurru:


e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -126,10 +129,14 @@ export default class Dashboards extends BasePage {
 
     let noDataPanels: string[] = [];
     let missingMetrics: string[] = [];
+    let missingMetricIndexes: number[] = [];
 
     for (let i = 0; i <= timeout; i += Timeouts.THIRTY_SECONDS) {
       noDataPanels = await this.elements.noDataPanelName.allTextContents();
-      missingMetrics = Array.from(noDataPanels).filter((e) => !noDataMetrics.includes(e));
+      missingMetricIndexes = noDataPanels.flatMap((name, index) =>
```
yurkovychv: Why this change is needed? What is wrong with the current logic?
@kiranvuyurru

yurkovychv (changes requested):

e2e_tests/pages/dashboards/dashboards.page.ts
kiranvuyurru: Error: locator.screenshot: Error: strict mode violation: locator('//section[contains(@data-testid, "RAM")]') resolved to 2 elements:
    1) <section tabindex="0" aria-labelledby="_r25_" class="css-1ham21n-panel-container" data-testid="data-testid Panel header Buffer Pool Size of Total RAM">…</section> aka getByTestId('data-testid Panel header Buffer Pool Size of Total RAM')
    2) <section tabindex="0" aria-labelledby="_r75_" class="css-1ham21n-panel-container" data-testid="data-testid Panel header RAM">…</section> aka getByTestId('data-testid Panel header RAM')

Call log:
  - waiting for locator('//*[@id="grafana-iframe"]').contentFrame().locator('//section[contains(@data-testid, "RAM")]') 
  - because of this issue

yurkovychv: Just fix it with the 'first()' function instead of updating the logic here

kiranvuyurru: if we use first() it will always pick the first panel. this only works if we have only 1 matching panel

yurkovychv: This is only neede to make screenshots which I believe is not useful since we attach trace
Lets revert the changes and remove the screenshoting

yurkovychv: @travagliad @peterSirotnak do you agree?

kiranvuyurru: will remove the whole screen shorting then

travagliad: Absolutely, don't know about you but trace is my go-to tool for debugging. I'll open a PR to also attach the reporter so we can navigate tracers easily, then screenshot wouldn't be needed.

yurkovychv (approved):

PR #998, opened 2026-06-05
PMM-14901 more RTA tests, cli tests fixes

yurkovychv:


cli/support/types/pipe-assertions.class.ts
```diff
@@ -29,14 +29,6 @@ class PipeAssertions {
     for (const val of expectedValues) {
       expect.soft(this.text, `Verify ${this.pipeName} contains '${val}'`).toContain(val);
     }
-    console.log(test.info().errors.length);
-    const errorMsg = test.info().errors.length === 0
-      ? ''
-      : ` But got ${test.info().errors.length} error(s):\n${this.getErrors()}`;
-    expect(
```
peterSirotnak, line 36: Why are we removing assertion?

yurkovychv: this assertion is not needed here
there is an assertion above

travagliad (approved):

PR #1013, opened 2026-06-12
PMM-7 bump checkout action version

yurkovychv:


.github/workflows/helm-tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
       steps:
         - name: Checkout code
-          uses: actions/checkout@v4
+          uses: actions/checkout@v6
```
ademidoff: We are supposed to pin the version, check here - https://github.com/percona/pmm/blob/v3/.github/workflows/main.yml#L41

ademidoff: Kind of security best practice )

yurkovychv: thanks

peterSirotnak (approved):

PR #1018, opened 2026-06-17
PMM-14931: Control ClickHouse configuration using environment variables

peterSirotnak:
https://github.com/percona/pmm-qa/actions/runs/27674657029

### DIFF
--- a/e2e_tests/tests/dockerConfiguration/clickhouseConfig.test.ts
+++ b/e2e_tests/tests/dockerConfiguration/clickhouseConfig.test.ts
@@ -1,0 +1,30 @@
+import pmmTest from '@fixtures/pmmTest';
+import dataTest from '@fixtures/dataTest';
+
+pmmTest.describe('PMM Tests to verify clickhouse configuration file', () => {
+  const dockerContainerName = 'pmm-server-clickhouse-config';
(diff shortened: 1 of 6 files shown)


e2e_tests/tests/dockerConfiguration/clickhouseConfig.test.ts
```diff
@@ -0,0 +1,51 @@
+import pmmTest from '@fixtures/pmmTest';
+import dataTest from '@fixtures/dataTest';
+import { expect } from '@playwright/test';
+
+pmmTest.describe('PMM Tests to verify clickhouse configuration file', () => {
+  pmmTest.describe.configure({ mode: 'serial' });
```
yurkovychv: I think we don't need to have `serial` mode here
is there a reason for it?

peterSirotnak: Yes we are using same docker container name for both images, and for running multiple docker images I found out that serial i more stable.

yurkovychv: why can't we use different names for docker containers?
in this mode `If one of the serial tests fails, all subsequent tests are skipped.`

peterSirotnak: Removed serial mode.

yurkovychv (approved):

PR #1110, opened 2026-07-24
fix(deps): resolve Dependabot security update failures on main

travagliad:
## Why

Five Dependabot security checks failed on `main` (commit c6c5c92a) because transitive dependencies in `codeceptjs-e2e` could not be auto-updated (`diff`, `joi`, `js-yaml`, `brace-expansion`) and `setuptools` was capped below the patched version.

Latest FB build (pmm-submodules PR 4490) is green — no product or test bug to fix there.

## How

- Add npm `overrides` in `codeceptjs-e2e/package.json` to pin patched versions of vulnerable transitive deps.
- Regenerate `package-lock.json` via `npm install`.
- Raise `setuptools` floor in `qa-integration/pmm_qa/requirements.txt` to `>=83.0.0` so pip security updates can resolve.

### DIFF
--- a/codeceptjs-e2e/package-lock.json
+++ b/codeceptjs-e2e/package-lock.json
@@ -941,19 +941,6 @@
         "node": ">=8"
       }
     },
-    "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.14.2",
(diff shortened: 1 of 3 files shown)


travagliad:
Closed per request.

PR #1125, opened 2026-07-30
PMM-15265 pmm framework migration

yurkovychv:
[nightly run](https://github.com/percona/pmm-qa/actions/runs/30538612833)
[e2e matrix run](https://github.com/percona/pmm-qa/actions/runs/30538088279)


qa-integration/pmm_qa/pmm-framework/lib/execution.sh
```diff
@@ -0,0 +1,290 @@
...
+      conflict="two $DB_TYPE setups"
+    elif [[ $DB_TYPE == PS || $DB_TYPE == MYSQL ]]; then
+      if [[ -n $mysql_data_owner ]]; then
+        conflict="$mysql_data_owner and $DB_TYPE setups (shared mysql_cluster_data and host ports)"
+      fi
+      mysql_data_owner=$DB_TYPE
+    fi
+    seen_types["$DB_TYPE"]=1
+  done
+
```
kiranvuyurru, line 93: shouldn't this handle conflict between pgsql and pdpgsql too

yurkovychv: Good catch, there is a clash with pdpgsql and pgsql with replication
added this case to the preflight check

qa-integration/pmm_qa/mysql/tasks/prepare_install_mysql.yml
```diff
@@ -5,9 +5,11 @@
 
 - name: Prepare socket folders
   shell: |
-    sudo rm -fr /tmp/mysql-sockets/{{ item }}
-    sudo mkdir -p /tmp/mysql-sockets/{{ item }}
-    sudo chmod 0777 /tmp/mysql-sockets/{{ item }}
+    docker run --rm -v /tmp:/host-tmp busybox:1.37.0 sh -c '
```
travagliad: Isn't better to have all versions coded in one place? so updates are easier

yurkovychv: this is just a tool to execute commands 
not sure if we will ever update the version here
but I can try moving it out if needed

qa-integration/pmm_qa/postgresql/postgresql-setup.yml
travagliad, line 115: changing this filename can break other setups we might not have looked into?

yurkovychv: the changes are tested on nightly and on matrix 
nothing is broken

qa-integration/pmm_qa/mysql/tasks/prepare_install_mysql.yml
yurkovychv: moved the image out on a higher level

travagliad (approved):

PR #1126, opened 2026-07-30
Skip PMM-T289 upgrade test

kiranvuyurru:
Mark the PMM-T289 upgrade test as skipped by changing Scenario to Scenario.skip in codeceptjs-e2e/tests/upgrade/upgradePMM_test.js. This prevents the 'Verify Whats New link is presented on Update Widget' test from running in E2E suites as the link is no longer present in the panel

### DIFF
--- a/codeceptjs-e2e/tests/upgrade/upgradePMM_test.js
+++ b/codeceptjs-e2e/tests/upgrade/upgradePMM_test.js
@@ -36,7 +36,7 @@ Before(async ({ I }) => {
   await I.Authorize();
 });
 
-Scenario(
+Scenario.skip(
   'PMM-T289 - Verify Whats New link is presented on Update Widget @pmm-upgrade',
   async ({ I, homePage }) => {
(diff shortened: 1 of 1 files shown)


codeceptjs-e2e/tests/upgrade/upgradePMM_test.js
```diff
@@ -36,7 +36,7 @@ Before(async ({ I }) => {
   await I.Authorize();
 });
 
-Scenario(
+Scenario.skip(
```
peterSirotnak, line 39: We still use this for next 5 versions upgrade 3.8.1 Still support upgrade no reason to skip this.

peterSirotnak (changes requested):

codeceptjs-e2e/tests/upgrade/upgradePMM_test.js
kiranvuyurru: ahh, i see. got it

PR #1157, opened 2026-08-10
Apply all outstanding Dependabot updates across cli, codeceptjs-e2e and e2e_tests

travagliad:
Replaces all 11 open Dependabot PRs with one consolidated, exercised update. Companion to #1156, which stops Dependabot opening one PR per dependency.

Supersedes: #1107, #1114, #1115, #1117, #1118, #1119, #1131, #1132, #1133, #1134, #1148 — all can be closed once this merges.

## Security

`npm audit` across the three packages goes from **30 vulnerabilities (2 critical, 18 high)** to **14 (0 critical)**.

| Package | Before | After |
|---------|--------|-------|
| `cli` | 12 (8 high, 4 moderate) | **0** |
| `e2e_tests` | 0 | **0** |
| `codeceptjs-e2e` | 18 (2 critical, 10 high) | 14 (0 critical) |

The remaining 14 all sit inside `codeceptjs@3.7.9`'s own tree — its bundled axios/joi/uuid/mocha and `@codeceptjs/detox-helper` → `react-native` → `metro` → `image-size`.

**These do not clear by moving to CodeceptJS 4.** That was measured, not assumed: `codeceptjs@4.1.0` installed on its own reports **11 vulnerabilities (1 low, 7 moderate, 3 high)**, including `undici` via a new `ai`/`@ai-sdk` dependency chain and `uuid <11.1.1`. CodeceptJS 4 trades one set of advisories for another rather than clearing them, so it is not a security argument for the migration.

## Two dependencies Dependabot got wrong

**`faker` 5.5.3 → 6.6.6 (#1107) would have broken the suite.** `faker@6.6.6` is the sabotaged January 2022 release — 3.8 KB unpacked, no working API. Seven test files `require('faker')`. Instead of taking that bump, the legacy package is removed and those files now use `@faker-js/faker`, which this package already depended on:

```
faker.datatype.number()      -> faker.number.int()
faker.random.alpha()         -> faker.string.alpha()
faker.random.alphaNumeric()  -> faker.string.alphanumeric()
faker.internet.userName()    -> faker.internet.username()
```

**`typescript` → 7.0.2 is not installable.** No `@typescript-eslint` release supports TypeScript 7 — every published version peers `typescript <6.1.0`.

## Dead dependencies removed

- **`request`** — unused, EOL since 2020, and the sole source of both critical advisories (`form-data <=2.5.5`, plus `qs` and `tough-cookie`), none of which had a fix available.
- **`mysql`** (v2) — unused; the tests use `mysql2`.
- **`faker`**, **`@types/faker`** — see above.

## Breaking upgrades and the code changes they required

**ESLint 9 → 10 + eslint-plugin-unicorn 55 → 73** (`e2e_tests`)
- unicorn 73's `filename-case` started checking directory names; kebab-case directories are this repo's convention, so the rule is restricted to filenames
- ESLint 10's new `no-useless-assignment` flagged a `noDataPanels` initialiser that is always overwritten — scoped it into the loop

**TypeScript 5 → 6** (`e2e_tests` only — see the held-back table)
- TS 6 resolves indexing on `typeof X & Record` through the index signature, so the valkey dashboard loop iterates typed keys instead of `for..in`
- TS 6 errors on deprecated `baseUrl`, `node10` resolution and the `es5` target. Changing any of them alters module/emit semantics, so they are silenced with `ignoreDeprecations: "6.0"` and left for a separate migration.

**eslint-plugin-playwright 0.18 → 2.11** (`cli`)
- its `valid-describe-callback` rule flags `async` `test.describe` callbacks. Playwright never awaits them, so the `async` keyword was removed — no behaviour change.

**TLS handling** (`codeceptjs-e2e`)
- `pr.codecept.js` set `NODE_TLS_REJECT_UNAUTHORIZED = '0'`, disabling certificate verification for the whole Node process and printing a warning on every run. That is now scoped to a per-call `https.Agent`, built inline at each of the four call sites that need it (`pr.codecept.js` REST `onRequest`, `custom_steps` `downloadZipFile`, `dumpAPI` `downloadDump`, `triggerAdvisors`).
- The REST helper's pre-existing top-level `httpsAgent` property was **dead config** — `lib/helper/REST.js` only builds an agent from its `httpAgent` option, which throws without `ca`/`key`/`cert`. REST worked purely because the environment variable disabled verification globally. The agent is now attached in `onRequest`, which is the supported hook. Verified against a self-signed server with the variable unset: no agent → `DEPTH_ZERO_SELF_SIGNED_CERT`, top-level `httpsAgent` → `DEPTH_ZERO_SELF_SIGNED_CERT`, `onRequest` agent → `200`.
- The agent sets `keepAlive: false`. An earlier revision of this description claimed that also cleared `MaxListenersExceededWarning: 11 error listeners added to [Socket]`; CI shows it does not. That warning is **pre-existing and unrelated to this PR** — the `@bm-mongo` job logs it 12 times on the #1156 baseline and 10 times here. Left alone deliberately; it deserves its own ticket (see follow-ups).

## Bonus: `cli`'s lint was silently dead

`npm run lint` had never actually linted anything in `cli`. `.eslintrc` pointed the TS parser at `./cli/tsconfig.json`, which does not resolve from inside `cli/`, so all 26 TS files failed with `Parsing error: Cannot read file` and no rule ever ran. Plain `.js` files were unparseable too, because only the `*.ts` override set `parserOptions`.

Fixing the path surfaced the real findings. Also corrected three root rules that had been left at ESLint defaults and contradict the style used everywhere in the package (`quotes` → single, `indent` → 2, `one-var` → never), removed two unused variables, and applied the auto-fixable formatting. `npm run lint` (eslint + tsc) now passes.

## What is deliberately held back

| Held | Version | Why |
|------|---------|-----|
| `codeceptjs` | 3.7.9 | CodeceptJS 4 across 215 test files / ~30k LOC is its own project, and it does **not** reduce the advisory count (see Security) |
| `mongodb` (codeceptjs-e2e) | 3.7.2 | Driver 4+ discovers the replica set and dials members by their advertised names (`rs101`/`rs102`/`rs103`), which do not resolve on the runner; `directConnection: true` avoids that but pins to a possibly-secondary PSS member, so `createUser` fails with `not primary`. Reaching driver 7 needs primary-aware connection handling — a migration, not a bump. `mongodb@3.7.4` has **0 advisories**, so holding it costs nothing on security. |
| `typescript` (cli, codeceptjs-e2e) | ~5.5.4 | `@typescript-eslint` 7.18.0's `typescript-estree` declares `SUPPORTED_TYPESCRIPT_VERSIONS = '>=4.7.4 <5.6.0'`. Moving the parser to 8.x fails with `ERESOLVE` because `eslint-config-airbnb-typescript` 18 peers `@typescript-eslint` 7. Note the tilde — `^5.5.4` resolves to 5.9.3, still out of range. |
| `typescript` (e2e_tests) | 6.0.3 | **Not** held — this package is already on `@typescript-eslint` 8.67.0, which supports `>=4.8.4 <6.1.0` |
| `eslint` (cli, codeceptjs-e2e) | 8.57.1 | `eslint-config-airbnb*` is unmaintained and peers eslint ≤8; `eslint-plugin-import` supports ≤9. ESLint 10 means replacing the whole airbnb config (`eslint-plugin-import-x` + `@stylistic`) — a separate change. `e2e_tests`, which already uses flat config, **is** on ESLint 10. |
| `@faker-js/faker` (codeceptjs-e2e) | 9.9.0 | v10 is ESM-only; this package is CommonJS |

## Verification

Everything below was run locally on this branch:

| Check | Result |
|-------|--------|
| `cli` — `npm run lint` (eslint + tsc) | passes (first time ever), no TS-version warning |
| `cli` — `playwright test --list` | 214 tests in 18 files |
| `cli` — `npm audit` | 0 vulnerabilities |
| `e2e_tests` — `tsc --noEmit` | passes |
| `e2e_tests` — `eslint .` | passes |
| `e2e_tests` — `playwright test --list` | 164 tests in 31 files |
| `e2e_tests` — `npm audit` | 0 vulnerabilities |
| `codeceptjs-e2e` — `codeceptjs dry-run -c pr.codecept.js` | exit 0, no load errors |
| `codeceptjs-e2e` — `eslint --ext .js tests/` | 19 errors — **identical to `main`**, pre-existing, untouched |
| `codeceptjs-e2e` — REST TLS probe vs self-signed server | `onRequest` agent returns 200 with `NODE_TLS_REJECT_UNAUTHORIZED` unset |

Local checks cannot execute the tests themselves — they need a live PMM Server. The `E2E tests Matrix` workflow that runs on this PR is the real gate, and it has now completed green: run [31430848032](https://github.com/percona/pmm-qa/actions/runs/31430848032) on `0f73d2e`, **all 32 jobs passing** — including `@bm-mongo`, `@pmm-ps-pxc-haproxy-integration`, and `@rbac` (which is flaky on `main`). No certificate or TLS errors in any job, and the `NODE_TLS_REJECT_UNAUTHORIZED` warning is gone (1 occurrence on the #1156 baseline, 0 here).

## Follow-ups worth their own tickets

1. CodeceptJS 3 → 4. Note this is **not** a security win — v4 carries 11 advisories of its own.
2. MongoDB driver 3.7 → 7, with replica-set-aware connection handling (or resolvable member hostnames in CI).
3. Replace the airbnb ESLint config in `cli` and `codeceptjs-e2e` so they can reach ESLint 10 and `@typescript-eslint` 8 — which in turn unblocks TypeScript 6 there.
4. Move `cli`/`e2e_tests` off `node10` module resolution so `ignoreDeprecations` can go and TypeScript 7 becomes reachable.
5. `MaxListenersExceededWarning` in the CodeceptJS suite — pre-existing, ~10-12 occurrences per `@bm-mongo` run on both this branch and `main`.
6. The 19 pre-existing lint errors in `codeceptjs-e2e/tests/` — all auto-fixable, left out to keep this diff about dependencies.


e2e_tests/tsconfig.json
```diff
@@ -3,6 +3,9 @@
     "target": "ESNext",
     "module": "CommonJS",
     "moduleResolution": "node",
+    // TypeScript 6 deprecates node10 resolution; moving off it needs extension-ful
+    // relative imports, so keep it here and migrate separately.
```
travagliad:
```suggestion
```

e2e_tests/eslint.config.mjs
```diff
@@ -232,7 +232,8 @@ export default defineConfig([
       // -----------------------------------------------------------------------
       // Plugin: Unicorn
       // -----------------------------------------------------------------------
-      'unicorn/filename-case': ['error', { case: 'camelCase' }],
+      // unicorn 73 started checking directory names too; this repo uses kebab-case dirs
```
travagliad:
```suggestion
```

codeceptjs-e2e/tests/helper/mongoDB.js
```diff
@@ -150,7 +142,8 @@ class MongoDBHelper extends Helper {
    * @returns {Promise<unknown>}
    */
   async mongoAddUser(username, password, roles = [{ db: 'admin', role: 'userAdminAnyDatabase' }]) {
-    return this.client.db().admin().addUser(username, password, { roles });
+    // Admin.addUser() was removed in driver 6; createUser is the supported equivalent
```
travagliad:
```suggestion
```

cli/tsconfig.json
```diff
@@ -8,6 +8,9 @@
     "esModuleInterop": true,
     "allowSyntheticDefaultImports": true,
     "baseUrl": ".",
+    // TypeScript 6 deprecates baseUrl, node10 resolution and the ES5 target;
+    // changing any of them alters module/emit semantics, so migrate separately.
```
travagliad:
```suggestion
```

codeceptjs-e2e/pr.codecept.js
```diff
@@ -74,9 +73,13 @@ exports.config = {
     REST: {
       endpoint: process.env.PMM_UI_URL || pmmUrl,
       timeout: 60000,
-      httpsAgent: new Agent({
-        rejectUnauthorized: false,
-      }),
+      // The helper only builds an agent from its own httpAgent option, and that
+      // option requires ca/key/cert - it throws without them. PMM Server's
+      // certificate is self-signed, so the agent is attached per request, which
+      // is the documented hook for exactly this.
```
travagliad:
```suggestion
```

codeceptjs-e2e/tests/helper/httpsAgent.js
```diff
@@ -0,0 +1,13 @@
...
+// to be relaxed for requests aimed at it. Scoping that to this agent keeps the
+// rest of the process on normal TLS rules, rather than turning verification off
+// for everything via NODE_TLS_REJECT_UNAUTHORIZED=0 — which Node reports as
+// "makes TLS connections and HTTPS requests insecure" on every run.
+//
+// keepAlive is off deliberately. A pooled socket collects one 'error' listener
+// per in-flight request, and once more than ten share a TLSSocket Node raises
+// MaxListenersExceededWarning. Against a local server the extra handshakes cost
+// far less than the noise.
+module.exports = new Agent({ rejectUnauthorized: false, keepAlive: false });
```
travagliad: Don't create "httpAgent.js" file, do it all directly in each method you changing for this, even though it's repeatable.

cli/tests/postgreSql.spec.ts
```diff
@@ -106,7 +107,7 @@ test.describe('PMM Client CLI tests for PostgreSQL Data Base', { tag: '@pgsql' }
     const hosts = (await cli.exec('sudo pmm-admin list | grep "PostgreSQL" | awk -F" " \'{print $3}\''))
       .getStdOutLines();
     let n = 1;
-    for (const host of hosts) {
+    for (let i = 0; i < hosts.length; i++) {
```
travagliad: Can we stick to the first one? or maybe do for each if it's not possible?

travagliad: Do the same wherever you find ++ syntax

travagliad:
Test nightly: https://github.com/percona/pmm-qa/actions/runs/31447197650/job/93643854438

travagliad:
That job's only failure is `PMM-T2036 - Verify MongoDB PBM dashboard` (89 passed / 1 failed / 21 skipped):

```
Expected 0 Elements without data but found 3 on Dashboard .../d/mongodb-backup-details/...
Report Names are Last Successful Backup,Backup Sizes,Backup Duration
    at Object.verifyThereAreNoGraphsWithoutData (tests/pages/dashboardPage.js:1339:20)
```

It's a pre-existing flake, not something this PR introduced:

- The same test failed the identical assertion on plain `main` nightlies [#239](https://github.com/percona/pmm-qa/actions/runs/31133537909) (08-07) and [#233](https://github.com/percona/pmm-qa/actions/runs/30724706204) (08-02), on different PMM server IPs, with none of this branch's code involved. It passed on #240–#242, #244, #237, #238.
- Those `main` failures ran **Chromium 127** (main's playwright 1.45), so the playwright 1.45 → 1.62 bump in this PR can't be the cause of a failure that already happens without it.
- Same night, main's own nightly [#244](https://github.com/percona/pmm-qa/actions/runs/31444815961) failed with the *same* helper on MongoDB Cluster Summary (`Expected 9 ... but found 11`) — while the `@nightly` group that failed there **passed** here on Chromium 151.

Across the rest of that run, 15 of 16 jobs are green, including all 14 setup jobs. No TLS/certificate errors, no module-load or `require` failures, no faker API errors — nothing pointing at the dependency changes.

The deltas are small (0→1 and 9→11 panels) on a `from=now-5m` window, which reads as metrics not yet scraped when the dashboard opens. 3 of the 5 failed `main` nightlies in the last 10 days are this same no-data assertion class; I've listed it as a follow-up in the PR description rather than chasing it here.

The required gate, `E2E tests Matrix` [run 31430848032](https://github.com/percona/pmm-qa/actions/runs/31430848032), is green on all 32 jobs.

---

yurkovychv (approved):

PR #1168, opened 2026-08-12
PMM-15307 initial ha tests

yurkovychv:


.github/workflows/ha-e2e-tests.yml
```diff
@@ -0,0 +1,220 @@
+---
+name: ha-e2e-tests
```
yurkovychv, line 2: Ignore the workflow code, this is just a boilerplate

e2e_tests/helpers/constants.ts
```diff
@@ -0,0 +1,2 @@
+/** Kubernetes namespace the PMM deployment runs in, used by `k8sHelper`. */
+export const k8sNamespace = 'pmm';
```
travagliad: This one looks unnecessary, I checked it's only used in the constructor, so we might as well hardcode it in there, I like the idea of having a constant files for anything hardcoded, but for something only mentioned there and nothing else I'm not sure if it brings value or add another layer

e2e_tests/tests/ha/haBadgeHealth.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      });
+
+      // Never reload: at 0 replicas nothing serves the page. The app's own 15s
+      // refetch drives every transition, Unreachable included.
+      await pmmTest.step(`Verify the HA badge is ${expectedHealth}`, async () => {
+        await expect(
+          highAvailabilityPage.elements.badge,
+          `${expectedNodes - replicas} of ${expectedNodes} nodes are down, so the HA badge must be ${expectedHealth}`,
+        ).toHaveText(expectedHealth, { timeout: Timeouts.FIVE_MINUTES });
+      });
```
travagliad: If we're using test.step for reporting purposes, this will add a spoiler that when you expand it, the message will be pretty much the same, I think I would remove the step around expect here as expect is already going to be a step in the report by default (I would keep it for the methods though, like you did in previous line)

e2e_tests/tests/ha/haBadgeHealth.test.ts
```diff
@@ -0,0 +1,95 @@
...
+let statefulSet: string | undefined;
+
+pmmTest.beforeEach(async ({ api, grafanaHelper, haClusterHelper }) => {
+  await grafanaHelper.authorize();
+  await haClusterHelper.ensureServing(api.haApi, expectedNodes);
+});
+
+pmmTest.afterEach(async ({ api, haClusterHelper }) => {
+  await haClusterHelper.ensureServing(api.haApi, expectedNodes);
+});
```
travagliad, line 23: I might be wrong here, as I'm not sure about the specifics of this, but it looks like we're duplicating "ensuring serving" by mixing afterEach + beforeEach, could you confirm?

e2e_tests/tests/postRelease.test.ts
```diff
@@ -69,6 +79,8 @@ pmmTest(
 pmmTest(
   'PMM-T2202 - Verify PMM new version on percona.com/downloads @post-release @downloads',
   async ({ downloadsPage }) => {
+    const expectedVersion = getExpectedVersion();
+
```
travagliad: I see this was breaking the run when this wasn't set, but can we make it beforeAll so it is executed once and not every test?

travagliad (changes requested):
Marking as changes requested so Claude don't bug me for review again :D

e2e_tests/tests/ha/haBadgeHealth.test.ts
yurkovychv: this was done on purpose
as the test itself scales down the cluster and after the test the cluster will be not operational 
so it is basically there for usability

e2e_tests/tests/postRelease.test.ts
yurkovychv: good point
done

e2e_tests/tests/ha/haBadgeHealth.test.ts
yurkovychv: done, thanks

e2e_tests/helpers/constants.ts
yurkovychv: done, thanks

travagliad (approved):

PR #1170, opened 2026-08-12
Validate RTA export behavior and empty state

kiranvuyurru:


e2e_tests/README.md
```diff
@@ -92,7 +92,6 @@ For test, page-object, fixture, and helper conventions, see [CONTRIBUTING.md](./
 - `@dashboards`
 - `@docker-configuration`
 - `@downloads`
-- `@fb-settings`
```
travagliad, line 95: Why this is being removed? It was removed from tags?

travagliad (approved):
Approved but please check the tag removal

e2e_tests/README.md
kiranvuyurru: yes, no test is using this tag in e2e_tests. And BTW the script automatically removed this.

PR #1195, opened 2026-08-18
Make e2e client/framework setup steps survive a hung command

travagliad:

- source: nightly CI run [32086846152](https://github.com/percona/pmm-qa/actions/runs/32086846152) (Nightly E2E tests Matrix, 2026-08-18)
- tests:
  - `.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` / `setup / valkey` — setup step "Setup PMM-Client against remote PMM Server" (job [95561056169](https://github.com/percona/pmm-qa/actions/runs/32086846152/job/95561056169))

## What failed

`setup / valkey` hung in `apt-get update` (no response from `azure.archive.ubuntu.com`, then `archive.ubuntu.com`) and never recovered. At exactly `timeout_minutes: 10` the step died — and because `fail-fast: true` on the setup matrix, all 12 other setup shards were cancelled and both `test execution / …` jobs failed waiting on setup. **One shard's transient network stall took down the entire 14-shard nightly.** No test ever ran.

The apt stall itself is external infra flake — nothing to fix here, and not reproducible on demand. What *is* ours is that the retry wrapping that step could not fire.

## Why the retry never fired

The step is wrapped in `nick-fields/retry@v3` with `max_attempts: 2`, yet the log shows only `##[group]Attempt 1` and then:

```
Error: kill EPERM
    at process.kill (node:internal/process/per_thread:277:13)
    at killPid (…/nick-fields/retry/v3/dist/index.js:1941:17)
    at killAll (…/nick-fields/retry/v3/dist/index.js:1915:27)
```

Two independent reasons, both fixed here:

1. **`retry_on: 'error'` excludes timeouts by construction.** The action's documented values are `any` (default) / `timeout` / `error`; `error` retries only a non-zero exit, never the action's own timeout. The one failure mode these setup steps are most exposed to — a network hang — was the one they could never retry.
2. **The `sudo` child makes the timeout path fatal.** On timeout the action kills the command with tree-kill *before* it evaluates `retry_on`. The command runs as `sudo bash -x pmm3-client-setup.sh`, so the process tree is root-owned and the runner-user Node process cannot signal it. The `EPERM` escapes as an uncaught exception and terminates the action outright — so fixing (1) alone would still not have produced a retry.

## The fix

Give each wrapped command an inner `timeout` that fires *before* the action's `timeout_minutes` (9m inside 10m, 19m inside 20m). A hang then self-terminates with exit 124, which the action sees as an ordinary command failure — no tree-kill, no `EPERM`, retry loop intact. `retry_on` is also widened to `any` so an action-level timeout is not silently unretryable either.

Applied to all four retry blocks that share this shape, in `runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` and `ha-e2e-tests.yml`.

## Verification

On a throwaway Linode VM (`perconalab/pmm-server:3-dev-latest`, pmm-admin 3.9.1):

- **Failure mode reproduced directly.** Killing the parent of a `sudo` command leaves a root-owned orphan, and the non-root caller gets `kill: Operation not permitted` — the `EPERM` the action dies on.
- **New shape behaves as intended.** `sudo timeout -k 30 <n> …` returns `EXIT_CODE=124` to the non-root caller with no orphan left behind.
- **Happy path unaffected.** Both wrapped commands still exit 0 as changed: `sudo timeout -k 30 9m bash -x pmm3-client-setup.sh …` → 0, and `timeout -k 30 19m ./pmm-framework/pmm-framework --database valkey` → `[1/1] valkey: OK`, exit 0.

What only a real CI run can confirm: that a genuine runner-side network stall now produces an Attempt 2 rather than a dead job. The apt hang cannot be induced on demand, so that part rests on the mechanism above, not on an observed retry.

## Not fixed here (separate, already tracked)

The nightly has *also* been red every night since 2026-08-11 for an unrelated reason, which tonight's run never reached because setup died first: `PMM-T1061` and `PMM-T1790` fail in `QueryAnalyticsData.searchByValue()` with `Clickable element "//input[contains(@name, "search")]" was not found`. That is a PMM product regression from PMM-14848 / percona/pmm#5537 — the QAN search input unmounts on its own debounced refetch — tracked as **PMM-15336** (In Progress). The tests are correct and are deliberately left untouched; relaxing them would hide that defect.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PXYonxvBpXSiCJkC2QuJNc)_

### DIFF
--- a/.github/workflows/ha-e2e-tests.yml
+++ b/.github/workflows/ha-e2e-tests.yml
@@ -147,10 +147,15 @@ jobs:
           timeout_minutes: 10
           retry_wait_seconds: '10'
           max_attempts: 2
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            sudo bash -x pmm3-client-setup.sh \
+            # Inner timeout must fire before timeout_minutes above: a step the
+            # action times out itself is killed via tree-kill, which cannot
+            # signal the root-owned `sudo` child and dies on EPERM, taking the
+            # whole retry loop with it. Self-terminating returns exit 124 and
+            # leaves the retry loop intact.
+            sudo timeout -k 30 9m bash -x pmm3-client-setup.sh \
               --pmm_server_ip "${SERVER_IP}" \
               --client_version "${PMM_CLIENT_VERSION}" \
               --admin_password "${ADMIN_PASSWORD}" \
@@ -163,11 +168,11 @@ jobs:
           timeout_minutes: 20
           retry_wait_seconds: '10'
           max_attempts: 2
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
             mkdir -m 777 -p /tmp/backup_data || true
-            ./pmm-framework/pmm-framework \
+            timeout -k 30 19m ./pmm-framework/pmm-framework \
               --parallel \
               --pmm-server-ip "${SERVER_IP}" \
               --pmm-server-password "${ADMIN_PASSWORD}" \
--- a/.github/workflows/ha-e2e-tests.yml
+++ b/.github/workflows/ha-e2e-tests.yml
@@ -150,11 +150,6 @@ jobs:
           retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            # Inner timeout must fire before timeout_minutes above: a step the
-            # action times out itself is killed via tree-kill, which cannot
-            # signal the root-owned `sudo` child and dies on EPERM, taking the
-            # whole retry loop with it. Self-terminating returns exit 124 and
-            # leaves the retry loop intact.
             sudo timeout -k 30 9m bash -x pmm3-client-setup.sh \
               --pmm_server_ip "${SERVER_IP}" \
               --client_version "${PMM_CLIENT_VERSION}" \
--- a/.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml
+++ b/.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml
@@ -108,10 +108,15 @@ jobs:
           timeout_minutes: 10
           retry_wait_seconds: '10'
           max_attempts: 2
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            sudo bash -x pmm3-client-setup.sh \
+            # Inner timeout must fire before timeout_minutes above: a step the
+            # action times out itself is killed via tree-kill, which cannot
+            # signal the root-owned `sudo` child and dies on EPERM, taking the
+            # whole retry loop with it. Self-terminating returns exit 124 and
+            # leaves the retry loop intact.
+            sudo timeout -k 30 9m bash -x pmm3-client-setup.sh \
               --pmm_server_ip "${SERVER_IP}" \
               --client_version "${{ env.PMM_CLIENT_VERSION }}" \
               --admin_password "${{ env.ADMIN_PASSWORD }}" \
@@ -123,11 +128,11 @@ jobs:
           timeout_minutes: 20
           retry_wait_seconds: '10'
           max_attempts: 2
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
             mkdir -m 777 -p /tmp/backup_data || true
-            ./pmm-framework/pmm-framework \
+            timeout -k 30 19m ./pmm-framework/pmm-framework \
               --parallel \
               --pmm-server-ip "${SERVER_IP}" \
               --pmm-server-password "${{ env.ADMIN_PASSWORD }}" \
--- a/.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml
+++ b/.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml
@@ -111,11 +111,6 @@ jobs:
           retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            # Inner timeout must fire before timeout_minutes above: a step the
-            # action times out itself is killed via tree-kill, which cannot
-            # signal the root-owned `sudo` child and dies on EPERM, taking the
-            # whole retry loop with it. Self-terminating returns exit 124 and
-            # leaves the retry loop intact.
             sudo timeout -k 30 9m bash -x pmm3-client-setup.sh \
               --pmm_server_ip "${SERVER_IP}" \
               --client_version "${{ env.PMM_CLIENT_VERSION }}" \


travagliad:
Second occurrence of this exact failure, same night, different shards — recorded here rather than as a duplicate PR.

Nightly run [32086631744](https://github.com/percona/pmm-qa/actions/runs/32086631744) (run #264, 01:00 UTC — three minutes before the #265 run this PR was opened from, same `main` sha `57611b1`):

- `setup / pxc` ([95560424753](https://github.com/percona/pmm-qa/actions/runs/32086631744/job/95560424753)) — **failed**, `Setup PMM-Client against remote PMM Server` hung in `apt-get update` and died at `timeout_minutes: 10` with the same `Error: kill EPERM` / `killPid` trace, only `##[group]Attempt 1` in the log.
- `setup / ps-gr` ([95560424710](https://github.com/percona/pmm-qa/actions/runs/32086631744/job/95560424710)) — hung identically, cancelled by fail-fast before its own timeout.
- Both stall at the same point: `azure.archive.ubuntu.com` `Ign:` on every suite, fallback to `archive.ubuntu.com`, then nothing after `Get:5 … jammy-security InRelease`.
- `test execution / @nightly` then failed on `Wait for all setup jobs to be ready`; no test ran.

So it isn't shard-specific — `valkey`, `pxc`, and `ps-gr` all hit it, which is consistent with the diagnosis here (the step's exposure, not any one database's setup).

## One gap this diff doesn't cover

In the same run, `test execution / @qan|@menu|@valkey-nightly|@permissions-nightly|@pt-summary-nightly|@pbm-nightly` ([95560424667](https://github.com/percona/pmm-qa/actions/runs/32086631744/job/95560424667)) hung in the *same* apt stall — inside `npx playwright install --with-deps`, which shells out to `apt-get` — and stalled after the identical `jammy-security InRelease` line. It sat there from 01:04:37 until the job's `timeout-minutes: 120` cancelled it at 03:04:30:

```
Terminate orphan process: pid (1909) (npm exec playwright install --with-deps)
```

That step is `Setup npm modules for e2e tests` in `runner-e2e-tests-codeceptjs-remote-nightly-tests.yml:122`, and it has no `nick-fields/retry` wrapper and no inner `timeout` — the 120-minute job timeout is the only bound in that file. One runner's network stall therefore burns two hours of a nightly instead of ten minutes, with no retry.

Same shape as the four blocks fixed here, so it may be worth folding into this PR:

```yaml
- name: Setup npm modules for e2e tests
  working-directory: pmm-qa/codeceptjs-e2e
  run: |
    timeout -k 30 15m npm ci
    timeout -k 30 15m npx playwright install --with-deps
    envsubst < env.list > env.generated.list
```

(No `sudo` in this one, so the `EPERM` half doesn't apply — but the unbounded-hang half does.)

Not opening a separate PR for it; investigator dedup stops on this one.


---

travagliad:
Recurrence + one gap this PR does not cover, from the next nightly run.

**Same failure again, different shard.** [Run 32091864890](https://github.com/percona/pmm-qa/actions/runs/32091864890) (run #266, 02:26 UTC) failed identically — this time on `setup / pxc` ([job 95575500069](https://github.com/percona/pmm-qa/actions/runs/32091864890/job/95575500069)), same step `Setup PMM-Client against remote PMM Server`, same signature:

```
Get:5 https://archive.ubuntu.com/ubuntu jammy-security InRelease [129 kB]   02:39:56
  … 9m22s of nothing …
Error: kill EPERM  at killPid (…/nick-fields/retry/v3/dist/index.js:1941:17)   02:49:18
```

Only `##[group]Attempt 1` in the log, `fail-fast` cancelled the other 13 shards, and `test execution / @qan|@menu|…` died in `Wait for all setup jobs to be ready`. [Run 32086631744](https://github.com/percona/pmm-qa/actions/runs/32086631744) (#264, 01:00 UTC) was the same thing on `setup / pxc` as well. So that's three shards across three runs in one night — the diagnosis in this PR holds, and the shard that draws the stall is arbitrary.

**Not covered here:** in the same run, `test execution / @nightly` ([job 95575499864](https://github.com/percona/pmm-qa/actions/runs/32091864890/job/95575499864)) hung in step `Setup npm modules for e2e tests` from 02:32:23 until the job cap at 04:32:05 — 2 hours, ignoring the cancel that landed at 02:49. Post-cleanup names the culprit:

```
Terminate orphan process: pid (1906) (npm exec playwright install --with-deps)
```

`playwright install --with-deps` shells out to apt, so it is the same stall class — but that step lives in `runner-e2e-tests-codeceptjs-remote-nightly-tests.yml` and is a bare `run:` with no `nick-fields/retry` wrapper and no inner `timeout`, so none of the four blocks changed here protect it. It burned ~2h of runner time without being what failed the run.

Worth folding an inner `timeout` into that step too, or leaving it for a follow-up — flagging rather than pushing to this PR, since it's outside the shape you fixed.

---

travagliad:
Recurred tonight, worse: [run 32156476603](https://github.com/percona/pmm-qa/actions/runs/32156476603) (`Nightly E2E tests Matrix`, same `main` @ `b8f6c3b`) died the same way — this time **three** setup shards hung in `apt-get update`, not one.

Identical signature in each: `azure.archive.ubuntu.com` `Ign`, fallback to `archive.ubuntu.com`, then a flat ~9-minute stall, then the step killed at `timeout_minutes: 10` with only `##[group]Attempt 1` ever logged and

```
Error: kill EPERM
    at killPid (…/nick-fields/retry/v3/dist/index.js:1941:17)
```

Affected shards: `pdpgsql` ([95774805833](https://github.com/percona/pmm-qa/actions/runs/32156476603/job/95774805833)), `psmdb-pss` ([95774806028](https://github.com/percona/pmm-qa/actions/runs/32156476603/job/95774806028)), `psmdb-sharded` ([95774805893](https://github.com/percona/pmm-qa/actions/runs/32156476603/job/95774805893)); `mysql` was cancelled mid-setup. The other 9 shards were cancelled by `fail-fast`, and both `test execution / …` jobs failed on `Wait for all setup jobs to be ready`. Again no test ran at all — `No files were found with the provided path: pmm-qa/codeceptjs-e2e/tests/output/result.xml`.

Second consecutive nightly lost to this, so no separate PR: this one covers it. Worth noting the retry budget is per-shard, so three simultaneous stalls would each need their own Attempt 2 to survive — the fix makes that possible where today it is structurally impossible.

---

travagliad:
Correcting my comment above — I said "this one covers it", and after pulling the full run history that claim is too strong. This PR makes the hang *retryable*; it does not stop the hang, and on the evidence a retry is roughly a coin flip rather than a fix.

**What the history actually shows.** I classified every `setup /` shard across 8 nightly runs by whether its apt could reach `azure.archive.ubuntu.com`:

| Run | Shards that lost the azure mirror | Of those, wedged to death |
|---|---|---|
| 08-10 → 08-17 (5 runs, 70 shards) | **0** | 0 |
| 08-18 01:03 | 6 / 14 | 1 (`valkey`) |
| 08-18 13:03 | 0 / 14 | 0 — **run passed** |
| 08-18 15:46 | 6 / 14 | 3 (`pdpgsql`, `psmdb-pss`, `psmdb-sharded`) |

Every "Setup PMM-Client" failure in the workflow's recorded history is on 2026-08-18. Before that, 70 consecutive shards fetched apt in 2–4s with zero `Ign` on the azure mirror.

**Not caused by anything we changed:** `runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` is unchanged since 15bc85a (2026-07-31), `qa-integration/pmm_qa/pmm3-client-setup.sh` since c5f3574 (2026-07-21), and the runner image is the same `ubuntu-22.04 / 20260729.566` on 08-16, 08-17 and every 08-18 run. The sole discriminator is mirror reachability: shards that reach `azure.archive.ubuntu.com` finish `apt-get update` in ~3s; shards that get `Ign` fall back to `https://archive.ubuntu.com`, and some of those never return.

**Why a retry is not the fix.** Attempt 2 runs on the same runner with the same broken route to the azure mirror. In the 01:03 run 4 of 5 shards that lost azure still completed via the HTTPS fallback, and in the 15:46 run all 3 that lost azure wedged — so a second attempt buys a real second chance, but it also turns a 10-minute failure into a 20-minute one when it doesn't, and with `fail-fast: true` on the setup matrix (nightly-e2e-tests-matrix.yml:83) one unlucky shard still takes down all 14.

I still think this PR is worth merging — the `EPERM` crash is a genuine bug and `retry_on: 'error'` genuinely can never retry a timeout. But it should land as hardening, not as the answer to "setup keeps failing". The cause worth fixing is that `apt-get update` has no timeout and no mirror resilience, so it can wedge for the full 10 minutes instead of failing in seconds.

---

.github/workflows/ha-e2e-tests.yml
```diff
@@ -147,10 +147,15 @@ jobs:
...
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            sudo bash -x pmm3-client-setup.sh \
+            # Inner timeout must fire before timeout_minutes above: a step the
+            # action times out itself is killed via tree-kill, which cannot
+            # signal the root-owned `sudo` child and dies on EPERM, taking the
+            # whole retry loop with it. Self-terminating returns exit 124 and
+            # leaves the retry loop intact.
```
travagliad: > 
```suggestion
```

.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml
```diff
@@ -108,10 +108,15 @@ jobs:
...
-          retry_on: 'error'
+          retry_on: 'any'
           command: |
             cd pmm-qa/qa-integration/pmm_qa
-            sudo bash -x pmm3-client-setup.sh \
+            # Inner timeout must fire before timeout_minutes above: a step the
+            # action times out itself is killed via tree-kill, which cannot
+            # signal the root-owned `sudo` child and dies on EPERM, taking the
+            # whole retry loop with it. Self-terminating returns exit 124 and
+            # leaves the retry loop intact.
```
travagliad:
```suggestion
```

yurkovychv (approved):

PR #1197, opened 2026-08-18
PMM-7 update clickhouse test, add-ha-test skill

yurkovychv:


.claude/skills/add-ha-test/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+
+After writing a file, re-read every comment and cut the ones that do not survive
+"could a competent reader work this out from the code?". The natural instinct is well past the bar.
+
+## Prove the test can fail
+
+Do not assume the test is working. A green HA test proves
+little on its own — a locator that resolves to nothing,
+or an oracle that echoes the thing under test, both look green. Mutate and
+re-run:
```
travagliad: Feels like something is missing here

"Rerun:" <something>

.claude/skills/add-ha-test/SKILL.md
```diff
@@ -0,0 +1,192 @@
+---
+name: add-ha-test
+description: >-
+  Write or debug Playwright e2e tests for PMM in High Availability mode —
+  anything under e2e_tests/tests/ha/ or tagged @pmm-ha. Use when automating a
+  PMM-Txxxx HA test case, when a test needs the Raft leader / failover /
+  leader-follower roles / pmm_ha_* metrics / the sidebar HA badge, when touching
+  k8s.helper.ts, haCluster.helper.ts, ha.api.ts, prometheus.api.ts or
+  pages/ha/**, when an @pmm-ha test fails.
```
travagliad: The description is loaded everytime we send a prompt, so we might want to make it more simple like this: 

I removed mention to failed pmm-ha test, I think this is something more suited into it's own triage skill

```suggestion
description: >-
  Write or debug Playwright e2e tests for PMM in High Availability mode — the
  test suite tagged @pmm-ha. Use when automating a PMM-Txxxx HA case, or 
  when touching the HA helpers, api or pages that drive the Raft leader, failover 
  and pmm_ha_* metrics.
```

.claude/skills/add-ha-test/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+| Asserting on | Use as truth |
+|---|---|
+| Sidebar badge, Inventory Nodes roles, anything rendering `/v1/ha/nodes` | `haClusterHelper.leaderFromPods()` |
+| `pmm_ha_leader_status` | `/v1/ha/nodes`, or the badge — different code paths off the same Raft state |
+| A failover actually happened | `haClusterHelper.lastPromotionTime()` before vs after, on the same pod |
+
+`leaderFromPods()` asks each pod `/v1/server/leaderHealthCheck` directly:
+**200 on the leader, 400 on followers**. It bypasses HAProxy, so it is
+independent of the aggregated API and of the LB. This is the same check
+HAProxy itself routes on, so it is a supported contract, not a scrape.
```
travagliad: This is also mentioned in cluster-facts, so I think we reference it and instead don't duplicate it here

```suggestion
`leaderFromPods()` asks each pod `/v1/server/leaderHealthCheck` directly, which
is the same check HAProxy routes on — a supported contract, not a scrape, and
independent of both the aggregated API and the LB. Status codes and the HAProxy
backend they come from: [references/cluster-facts.md](references/cluster-facts.md).
```

.claude/skills/add-ha-test/SKILL.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Prove the test can fail
+
+Do not assume the test is working. A green HA test proves
+little on its own — a locator that resolves to nothing,
+or an oracle that echoes the thing under test, both look green. Mutate and
+re-run:
+
+Invert the expectation, not the assertion — inverting `toEqual` to
+`not.toEqual` passes for the wrong reason.
```
travagliad:
```suggestion
## Prove the test can fail

A green HA test proves little on its own — a locator that resolves to nothing,
or an oracle that echoes the thing under test, both look green. Mutate and
re-run:

1. Change the expected **value**, not the matcher: assert the leader is a pod
   you know is not leading, or the Raft term is a number it is not.
2. Re-run and read the failure. It has to fail on your changed value — a
   timeout or a missing locator means the test was never asserting what you
   thought.
3. Revert.

Inverting the matcher (`toEqual` → `not.toEqual`) is not a mutation: anything
other than the expected value satisfies it, so it passes for the wrong reason.

**Done when** you have seen the test red for the right reason, then green again.```

yurkovychv:
@travagliad thanks for the inputs
applied

travagliad (approved):

PR #1216, opened 2026-08-20
Add skill gardener and verification depth skills

kiranvuyurru:
Adds two Claude skills:
Skill Gardener: Captures reusable lessons from completed work and applies approved improvements to skills and agent instructions.
Verification Depth: Ensures PMM testing uses direct, current, and appropriately timed evidence from APIs, CLI, logs, metrics, or stored data.


.gitignore
```diff
@@ -162,6 +162,8 @@ dist
 !.claude/integrations/
 !.claude/integrations/**
 !.claude/settings.json
+!.claude/skill-lessons.md
+!.claude/skill-gardener-state.json
```
travagliad, line 166: If we don't commit the lessons, how are we going to use it for autonomous runs?

.claude/skills/test-scope/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: test-scope
-description: Decide what a PMM change actually requires testing — which deployment mode (single-server Docker vs HA), and any extra dimensions (upgrade path, DB/version matrix, RBAC, backups). Use during QA planning, before provisioning, to pick the right environment and avoid both under-testing a regression and over-provisioning for a change that doesn't need it.
+description: "Decide PMM QA scope before provisioning: deployment mode, upgrade/DB/RBAC/backup dimensions, and causally linked regression checks. Use for planning what to exercise; do not use to judge whether direct evidence is deep enough."
```
travagliad: This skill is not meant to be used only when deciding pmm QA Scope before provisioning, It can be used locally where no provision Will take place too.

 This description might make it difficult for AI to trigger It. Is there a reason for changing the description such as we will have another skill for doing the same? If yes we should Merge both

.gitignore
kiranvuyurru: it is committed, "!" negates the ignore

.claude/skills/test-scope/SKILL.md
kiranvuyurru: you are right, i will make some changes

.gitignore
travagliad: Ouch, this is what happens when you review on your phone! 😆 Didn't noticed, thanks!

.claude/skills/test-scope/SKILL.md
```diff
@@ -1,17 +1,17 @@
...
+Read this when planning QA. It turns a ticket + diff into a short list of what to actually exercise, and which environment or provisioning skill each item implies. The goal is to catch the dimensions a plain single-server run would miss, without spinning up expensive environments a change doesn't need.
 
 Inputs you already have by this point: the Jira ticket (summary, AC, labels/components, dev comments) and the `percona/pmm` / `percona/grafana` diff (`git-diff` skill).
 
 ## Default
 
-Unless a dimension below applies, the change is tested on the **default single-server Docker** deployment — `linode-docker-provisioning`. Most fixes (one exporter metric, a dashboard panel, a CLI flag, a UI page with no clustered state) live here and need nothing more.
+Unless a dimension below applies, test on a suitable local or existing **single-server Docker** deployment, or create one with `linode-docker-provisioning`. Most fixes (one exporter metric, a dashboard panel, a CLI flag, a UI page with no clustered state) live here and need nothing more.
```
travagliad:
```suggestion
Unless a dimension below applies, the change is tested on the **default single-server Docker** deployment — `linode-docker-provisioning`. Most fixes (one exporter metric, a dashboard panel, a CLI flag, a UI page with no clustered state) live here and need nothing more.
```

We can't use local docker on VMs because Claude has a lot of restrictions over what this local instances can do, so this will open up the agents to have a lot of issues with permissions and routing to PMM while running tests, even opening PMM on browser sometimes fails.

travagliad (approved):

PR #1094, opened 2026-07-21
migrate(leftNavigation): PMM-T1830 server diagnostics logs to Playwright

kiranvuyurru:
## Summary

Migrates the only active scenario from `codeceptjs-e2e/tests/leftNavigation_test.js` into `e2e_tests/tests/helpCenter.test.ts`.

- **Scenario:** PMM-T1830 — Verify downloading server diagnostics logs
- **Execution tag:** `@new-navigation` (destination file tag; reuses existing `new-navigation` Playwright job)
- **Setup:** B1 — no DB provisioning

## Changes

- Append PMM-T1830 to `helpCenter.test.ts` with inline zip assertions via `readZipArchive`
- Add `e2e_tests/helpers/archive.helper.ts`
- Add `ServerApi.getPmmVersion()` and `/v1/version` endpoint
- Retire `leftNavigation_test.js` → `leftNavigation_migrated.js`
- Remove `@menu` from nightly CodeceptJS grep (no active CodeceptJS `@menu` scenarios remain)

## Excluded (per tracker)

- Commented `Data(...)` PMM-T433/T591
- `xScenario` PMM-T1051

## Validation

- ESLint on changed files: pass
- `CLEAN_ENVIRONMENT=false ./.cursor/scripts/run-migration-single-test.sh tests/helpCenter.test.ts '' false --grep PMM-T1830` → pass
- Full `helpCenter.test.ts` regression (12 tests) → pass

## Control branch follow-up

Tracker and `e2e_tests/graphify-out/` update will land on `CodeceptJS-migration` after this PR is opened.

<div><a href="https://cursor.com/agents/bc-35287f5d-e0b3-46e2-abc1-132c9cda2547"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-35287f5d-e0b3-46e2-abc1-132c9cda2547"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #1218, opened 2026-08-20
migrate(settings): serverLogs CodeceptJS to Playwright

kiranvuyurru:
## Migration

- Source: `codeceptjs-e2e/tests/serverLogs_test.js` (3 scenarios, tag `@fb-settings`, `BeforeSuite` warmup)
- Target: `e2e_tests/tests/serverLogs.test.ts` (new file)
- Migrated scenarios / preserved tags: PMM-T1902, PMM-T1903, PMM-T1904 — all keep `@fb-settings`
- Source and target files queried from existing graphs: `e2e_tests/graphify-out/graph.json`, `codeceptjs-e2e/graphify-out/graph.json` (during writer/reviewer; not regenerated by this branch)
- Setup used: `--database mysql`

## Environment requirement

These tests assert on `pmm-managed.log` line counts, so they need the server running at debug verbosity (`PMM_DEBUG=1`). Without it, PMM-T1902 and PMM-T1904 fail on a log line-count shortfall.

CI coverage for this comes from `e2e_tests/docker-compose.yml`, which sets `PMM_DEBUG=1` and is what `runner-e2e-tests-playwright.yml` brings the server up with — so any job routed through that runner (PR-CI, FB) is covered with no extra wiring.

The gap is **local provisioning**, which does not set it by default:

```
node provisioning/setup.ts --database mysql --server-env PMM_DEBUG=1
```

With `PMM_DEBUG=1`, all 3 scenarios pass reproducibly.

## Workflow coverage

These tests run in exactly the two places that supply `PMM_DEBUG=1`, and deliberately nowhere else:

- **FB** (`fb-e2e-suite.yml`): a Playwright `settings` job with the `@fb-settings` flag, mirroring the `alerting` job already in that file. Required because retiring the source off CodeceptJS's `_test.js` discovery glob removed PMM-T1902-1904 from the existing `@fb-alerting|@fb-settings` CodeceptJS job — which still runs green for `configuration/permissions_test.js` and `configuration/verifyPMMSettingsPageFunctionality_test.js`, so the loss would have been silent.
- **Playwright PR-CI** (`e2e-tests-matrix.yml`): a `server-logs` job (mysql setup, `@fb-settings` flag) — precedent: row 1's `leftNavigation`/`@menu` addition to the same file.
- **Nightly: intentionally excluded.** These scenarios carry only `@fb-settings` — no `@nightly` — and the nightly CodeceptJS buckets are `@nightly` and `@qan|@valkey-nightly|@permissions-nightly|@pt-summary-nightly|@pbm-nightly`, so they never ran there; adding them would have been new coverage, not preserved coverage, so the tag-bucket convention in that file does not call for it. It is also the one environment that cannot supply `PMM_DEBUG` (the remote nightly takes an external `pmm_server_address` with no compose and no server env), and that server is shared (`expected_setup_jobs: 14`) while this test deliberately floods its log past 50k lines.
- **CodeceptJS:** existing `@fb-settings` jobs and greps left unchanged.

## Review fixes applied

- **FB coverage restored** — the Playwright `@fb-settings` job described above.
- **Nightly entry dropped** — `test_execution_playwright` is back to `'@menu'`, byte-identical with `main`.
- **`getFileLineCount` counts logical lines** — `split('\n')` on newline-terminated content yields a trailing empty segment, so a 50,000-line log reported 50,001 and the helper name did not match what it returned. The trailing element is now dropped and the three expectations moved to `50_000` / `10` / `> 50_000`, matching `maxLogReadLines = 50000` in pmm's `managed/services/server/logs.go`. The pass/fail boundary is unchanged (`N+1 > 50_001` ⟺ `N > 50_000`).
- **Warmup kept to one worker** — `playwright.config.ts` sets `fullyParallel: true` and takes workers from `WORKERS`, so `beforeAll` ran once per worker and the 10,000-request warmup repeated per scenario. `describe.configure({ mode: 'default' })` pins the file to one worker. CI pins `workers: 1` everywhere, so this was latent there but live for local runs with `WORKERS` above 1.

Not changed, for the record: `ignoreHTTPSErrors: true` in the warmup context is **not** redundant with `playwright.config.ts`. `request.newContext()` from `@playwright/test` is the standalone `APIRequest` and does not inherit `use`, so removing it would break the warmup against an `https://` `PMM_UI_URL`.

## Validation / review results

- Static validation: `eslint`, `tsc --noEmit`, `actionlint`, `yamllint`, `generate_readme.py --check` — all clean
- MCP locator verification: N/A — no new/changed UI locators in this migration
- `getFileLineCount` verified directly against a built zip: a 10-line entry returns `10` whether or not it is newline-terminated (previously `11` and `10`), and an empty entry returns `0` rather than `1`
- Worker guard measured at `WORKERS=3`: the warmup ran 3 times across 3 worker pids without it and once with it
- Execution: `npx playwright test tests/serverLogs.test.ts` (full-file run) — PMM-T1902, PMM-T1903, PMM-T1904 all passed

## Source retirement

`codeceptjs-e2e/tests/serverLogs_test.js` renamed to `codeceptjs-e2e/tests/serverLogs_migrated.js` (no longer matches CodeceptJS test discovery's `_test.js` glob).


kiranvuyurru:
GitHub Actions: https://github.com/percona/pmm-qa/actions/runs/32408601223

.github/workflows/e2e-tests-matrix.yml
```diff
@@ -194,6 +194,18 @@ jobs:
...
+    name: Server logs tests
+    uses: ./.github/workflows/runner-e2e-tests-playwright.yml
+    secrets:
+      LAUNCHABLE_TOKEN: ${{ secrets.LAUNCHABLE_TOKEN }}
+    with:
+      pmm_server_version: ${{ inputs.pmm_server_image || github.event.inputs.pmm_server_image || 'perconalab/pmm-server:3-dev-latest' }}
+      pmm_client_version: ${{ inputs.pmm_client_version || github.event.inputs.pmm_client_version || 'latest-tarball' }}
+      pmm_qa_branch: ${{ github.event_name == 'pull_request' && github.head_ref || (inputs.pmm_qa_branch || github.event.inputs.pmm_qa_branch || 'main') }}
+      setup_services: '--database mysql'
+      pmm_test_flag: '@fb-settings'
```
travagliad: 🔴 **These three tests stop running in FB.**

`fb-e2e-suite.yml:167` greps `@fb-alerting|@fb-settings` on the **CodeceptJS** runner. Renaming the source off the `_test.js` glob removes T1902-1904 from it, and no Playwright `@fb-settings` job replaces them. The `@fb-` tag now runs everywhere except FB.

Precedent to copy is in the same file: `fb-e2e-suite.yml:170` already calls `runner-e2e-tests-playwright.yml` for `@alerting`. Add the equivalent for `@fb-settings`.

---

.github/workflows/nightly-e2e-tests-matrix.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - tags_for_tests: '@menu'
+          - tags_for_tests: '@menu|@fb-settings'
```
travagliad: 🔴 **This fires 10,000 requests at the shared nightly server.**

`serverLogs.test.ts:15` loops 10k sequential calls to inflate `pmm-managed.log` past 50k lines. The loop is inherited verbatim from the CodeceptJS original — fine there, since it only ever ran against a throwaway local server.

The nightly server is Jenkins-managed and shared with 14 setup shards plus two other test jobs, all asserting against it. Deliberately flooding its log there is new. Suggest keeping `@fb-settings` out of the nightly matrix and running it in FB + PR-CI only.

---

e2e_tests/tests/serverLogs.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  await requestContext.dispose();
+});
+
+pmmTest(
+  'PMM-T1902 - Verify no line_count parameter when downloading logs @fb-settings',
+  async ({ request }) => {
+    const response = await request.get(apiEndpoints.server.logs, { headers: GrafanaHelper.getAuthHeader() });
+    const actualLineCount = getFileLineCount(await response.body(), fileNameToCheck);
+
+    expect(actualLineCount, `File ${fileNameToCheck} has ${actualLineCount} lines, but expected 50001`).toBe(
```
travagliad: 🔴 **The `PMM_DEBUG` note in the body is wrong, and the real gap is the nightly.**

`e2e_tests/docker-compose.yml:21` already sets `PMM_DEBUG=1`, and `runner-e2e-tests-playwright.yml` brings the server up with that compose — so PR-CI is covered, contrary to "not currently a documented default anywhere else".

The remote nightly has no compose and nothing here sets the server env, so this exact-equality assertion hits the shortfall the body describes. Please correct the body and drop the nightly entry, or state where nightly gets `PMM_DEBUG`.

---

e2e_tests/helpers/archive.helper.ts
```diff
@@ -5,3 +5,14 @@ export const readZipArchive = (filepath: string): string[] => {
 
   return zip.getEntries().map(({ entryName }) => entryName);
 };
+
+export const getFileLineCount = (source: string | Buffer, fileName: string): number => {
```
travagliad, line 9: 🟡 **This file and the nightly matrix entry both exist in #1094 — whichever merges second conflicts.**

The body is honest about the duplication, which helps, but git still conflicts on `archive.helper.ts` and on `test_execution_playwright`. Nothing resolves that automatically once both PRs are open.

The repo's convention for depends-on-unmerged is draft + a linked dependency (`Depends on <url>`). Either mark this draft and stack it on #1094's branch, or merge #1094 first and rebase this one.

---

e2e_tests/api/backups.api.ts
```diff
@@ -52,8 +52,20 @@ export default class BackupsApi {
     await this.waitForSuccessfulBackupArtifact(service.service_name, scheduleName);
   };
 
+  getLocationsList = async () => {
```
travagliad: 🔵 The rename makes the diff bigger than the change. `getLocations` only needed to stop being `private`; renaming it to `getLocationsList` moves it out of the private block and rewrites the call site too.

Its only external caller is the log-warmup loop, which does not care about backups — it just needs an authenticated endpoint that logs. Faithful to the CodeceptJS original, so not worth blocking, but dropping `private` in place would have been a one-line diff.

---

e2e_tests/tests/serverLogs.test.ts
```diff
@@ -0,0 +1,59 @@
+import pmmTest from '@fixtures/pmmTest';
+import { expect, request as playwrightRequest } from '@playwright/test';
+import BackupsApi from '@api/backups.api';
+import GrafanaHelper from '@helpers/grafana.helper';
+import apiEndpoints from '@helpers/apiEndpoints';
+import { getFileLineCount } from '@helpers/archive.helper';
+import { pmmUrl } from '../playwright.config';
```
travagliad, line 7: 🔵 Two small things here.

`import { pmmUrl } from '../playwright.config'` is the only relative import in the file, in a suite that uses path aliases everywhere; and `ignoreHTTPSErrors: true` on line 12 just repeats what the config already sets.

Also, the three tests differ only in query string and expected count — `for (const c of cases) pmmTest(...)` would keep them as three separate tests (correct) without repeating the body three times.

---

travagliad (commented):
Reviewed with the `qa-code-review` skill. **3 🔴 · 1 🟡 · 2 🔵**

Mechanical pass first, all clean on this head — `npx eslint .` 0, `npx tsc --noEmit` 0, `actionlint` 0, `generate_readme.py --check` up to date. So nothing below is a style nit.

The migration itself is faithful: the 10k warmup, the `50001` expectation (= `maxLogReadLines` 50000 in pmm's `logs.go:45`, +1 from `split('\n')`), and the three test IDs all match the CodeceptJS source. Three separate tests rather than a loop inside one is the right shape.

What I would fix before merge is all about **where these now run**:

1. 🔴 FB loses T1902-1904 — the CodeceptJS grep is the only `@fb-settings` job, and the source is retired.
2. 🔴 The nightly entry points a 10k-request log flood at the shared Jenkins server.
3. 🔴 The `PMM_DEBUG` claim in the body is inverted: PR-CI already has it, the nightly does not.

🟡 is the #1094 overlap, 🔵 are cosmetic.

---

e2e_tests/helpers/archive.helper.ts
```diff
@@ -5,3 +5,14 @@ export const readZipArchive = (filepath: string): string[] => {
...
+  const zip = new AdmZip(source);
+  const entry = zip.getEntry(fileName);
+
+  if (!entry) {
+    throw new Error(`File ${fileName} not found in the ZIP`);
+  }
+
+  return entry.getData().toString('utf8').split('\n').length;
```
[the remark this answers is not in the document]
travagliad: 🟡 Unanswered bot finding — please answer and resolve it either way.

It is right that `getFileLineCount` returns logical+1, so the name lies. But 50001/11 is exactly what the CodeceptJS source asserted, and matches `maxLogReadLines = 50000` in pmm's `logs.go:45` plus the trailing split — so flipping to 50000/10 is not obviously better.

Either strip the trailing empty element and adjust all three expectations, or rename the helper. Author's call; it just needs one.

---

e2e_tests/tests/serverLogs.test.ts
```diff
@@ -0,0 +1,59 @@
...
+pmmTest.beforeAll(async () => {
+  const requestContext = await playwrightRequest.newContext({ baseURL: pmmUrl, ignoreHTTPSErrors: true });
+  const backupsApi = new BackupsApi(requestContext);
+
+  for (let i = 0; i < 10_000; i++) {
+    await backupsApi.getLocationsList();
+  }
+
+  await requestContext.dispose();
+});
```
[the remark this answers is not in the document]
travagliad: 🟡 Unanswered bot finding — and this one is valid, just latent.

`playwright.config.ts:10` sets `fullyParallel: true` and `:42` reads `WORKERS`. CI passes `1` today, so it is 10k requests and not 30k — but locally, or the day someone raises `WORKERS`, it triples.

It also compounds the shared-server concern on `nightly-e2e-tests-matrix.yml:173`: three workers × 10k against the Jenkins-managed server. Worth fixing even though CI currently hides it.

---

e2e_tests/helpers/archive.helper.ts
travagliad: Withdrawing this — I was wrong.

#1094 has merged. `origin/main` is at `4eb6d48` and already carries `archive.helper.ts` with `readZipArchive`, plus PMM-T1830 in `helpCenter.test.ts`. This branch's merge-base *is* that commit, so the diff only adds the 11 lines of `getFileLineCount` — no duplicated file, nothing to conflict on.

The PR body's "Known, accepted overlap" section is stale and can be dropped.

---

travagliad:
Second pass with `qa-code-review`, now including the bot-sweep, methods-and-comments rules.

**Correction first:** my 🟡 about duplicating #1094 was wrong and is withdrawn/resolved. #1094 merged — this branch's merge-base is `4eb6d48`, so the diff adds only the 11 lines of `getFileLineCount`. The body's "Known, accepted overlap" section is stale.

**Two CodeRabbit findings were open and unanswered** — replied on both. The worker/`beforeAll` one is valid and latent (`fullyParallel: true` + `WORKERS`), and it compounds the shared-nightly concern.

**Clean on the new checks:** no methods declared in the test file; zero code comments added; `getFileLineCount` has three callers, so it earns being a helper; and dropping `private` from `getLocations` is the direction we want, not a finding.

**Still standing:** the two 🔴 about *where* these run — FB loses T1902-1904, and the nightly entry points the log flood at the shared server. Both unchanged by the correction above.

Linters re-checked on `5a78f12`: eslint 0, `tsc --noEmit` 0, actionlint 0, `generate_readme.py --check` up to date.

---

.github/workflows/e2e-tests-matrix.yml
```diff
@@ -194,6 +194,18 @@ jobs:
       setup_services: '--database ps=8.4 --database psmdb --database pdpgsql'
       pmm_test_flag: '@LBAC'
 
+  server-logs:
```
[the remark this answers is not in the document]
kiranvuyurru: Fixed in 8a786de — the `server-logs` job is gone, and `e2e-tests-matrix.yml` is now byte-identical with `main`. Verified the premise first: `fb_tests` at :209 has no `if:`, so `fb-e2e-suite.yml`'s `settings` job already runs `@fb-settings` in every PR-CI run. The `alerting` precedent holds — `@alerting` appears only in `fb-e2e-suite.yml:180`, while `@menu` appears only in `e2e-tests-matrix.yml:159`, so neither doubles up.

---

e2e_tests/tests/serverLogs.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+pmmTest(
+  'PMM-T1902 - Verify no line_count parameter when downloading logs @fb-settings',
+  async ({ request }) => {
+    const response = await request.get(apiEndpoints.server.logs, { headers: GrafanaHelper.getAuthHeader() });
```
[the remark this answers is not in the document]
kiranvuyurru: Fixed in 8a786de — each scenario now asserts `expect(response.status(), response.statusText()).toEqual(200)` before the body reaches `getFileLineCount`, matching the pattern in `api/backups.api.ts:59`. A 401/500 or an HTML error page now fails on the status line, not as `Invalid or unsupported zip format`.

---

e2e_tests/tests/serverLogs.test.ts
kiranvuyurru: Fixed in 8a786de — the three scenarios now come from one `scenarios` table (`query`, `expectation`, `assertLineCount`) driving a single `pmmTest` body. Still three separate tests with byte-identical titles: `npx playwright test tests/serverLogs.test.ts --list --grep "@fb-settings"` reports T1902/T1903/T1904 as before, so the Zephyr keys and the tag grep are unaffected.

---

e2e_tests/api/backups.api.ts
kiranvuyurru: Taken, with one correction to the remedy. Dropping `private` in place does not lint: `eslint.config` has both `@typescript-eslint/member-ordering` and `perfectionist/sort-classes`, and leaving it where it was fails with

```
115:3  error  Member getLocations should be declared before all private method definitions
115:3  error  Expected "getLocations" (function-property) to come before "getArtifacts" (private-function-property)
```

So the move is forced; only the rename was not. 8a786de keeps the original `getLocations` name and moves it above the private block. `ensureLocalStorageLocation`'s call site is now unchanged, and the diff on this file is +10/-10 (one relocated block) instead of the rename churn.

---

.github/workflows/e2e-tests-matrix.yml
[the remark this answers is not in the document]
kiranvuyurru: Reverted in d442699 — the `server-logs` job stays. Author's call: dedicated PR-CI coverage for `@fb-settings` is worth the second run rather than depending solely on the `settings` job reached through `fb_tests`.

The observation above still stands factually (both jobs do run in a PR-CI build), so recording it as a deliberate trade-off rather than an oversight. The other three fixes from 8a786de are unchanged.

---

kiranvuyurru:
## CI: `@pmm-ps-integration` failed on `d442699` — setup, not this PR

`FB E2E tests / Percona Server for MySQL UI integration tests / e2e tests: @pmm-ps-integration` ([job](https://github.com/percona/pmm-qa/actions/runs/33641582117/job/100286011472)) went red, but it died in environment setup before any test body ran.

The Ansible task `Install tarball PMM client version` (`qa-integration/pmm_qa/tasks/install_pmm_client.yml:177`) failed after 3m47s fetching the client tarball from S3:

```
wget -O /pmm-client.tar.gz "https://pmm-build-cache.s3.us-east-2.amazonaws.com/PR-BUILDS/pmm-client/pmm-client-latest.tar.gz"
Connecting to pmm-build-cache.s3.us-east-2.amazonaws.com|3.5.133.194|:443... connected.
Unable to establish SSL connection.
```

`rc=4` → `ERROR: percona_server_for_mysql/percona-server-setup.yml playbook execution failed.` The `launchable record tests` step then failed with `FileNotFoundError: 'output/results.json'`, which is a consequence of the run producing no report, not a second problem.

**Why it is not this PR's:** the diff touches `e2e_tests/tests/serverLogs.test.ts`, `helpers/archive.helper.ts`, `helpers/apiEndpoints.ts`, `api/backups.api.ts`, the `@fb-settings` workflow entries and the README. It changes nothing in `qa-integration`, nothing about PMM client installation, and nothing reachable from `@pmm-ps-integration`. The failure is a TLS/network error against an external S3 endpoint. `Lint` passed on this same head.

**Next:** no fix to port — there is no code change that makes an unreachable S3 endpoint reachable. This qualifies for the one re-run (died before any test body ran). `rerun-failed-jobs` currently returns `403 This workflow is already running`, so the re-run happens once the run finishes; I'm keeping the PR watched until then.

---

kiranvuyurru:
Re-run passed — the S3/TLS failure was transient, as diagnosed above.

Run [33641582117](https://github.com/percona/pmm-qa/actions/runs/33641582117) attempt 2 completed `success` at 15:56Z. `@pmm-ps-integration` fetched the pmm-client tarball fine this time and its tests ran green, confirming the earlier red was the S3 endpoint and not this PR.

**`d442699` is green across the board** — including both `@fb-settings` jobs (`Server logs tests` and `FB E2E tests / Settings tests`) and `Lint`. No merge conflict; the PR is waiting on review.

---

.github/workflows/e2e-tests-matrix.yml
```diff
@@ -194,6 +194,18 @@ jobs:
...
+    uses: ./.github/workflows/runner-e2e-tests-playwright.yml
+    secrets:
+      LAUNCHABLE_TOKEN: ${{ secrets.LAUNCHABLE_TOKEN }}
+    with:
+      pmm_server_version: ${{ inputs.pmm_server_image || github.event.inputs.pmm_server_image || 'perconalab/pmm-server:3-dev-latest' }}
+      pmm_client_version: ${{ inputs.pmm_client_version || github.event.inputs.pmm_client_version || 'latest-tarball' }}
+      pmm_qa_branch: ${{ github.event_name == 'pull_request' && github.head_ref || (inputs.pmm_qa_branch || github.event.inputs.pmm_qa_branch || 'main') }}
+      setup_services: '--database mysql'
+      pmm_test_flag: '@fb-settings'
+
```
travagliad: This is duplicating the tests on e2e execution, we already execute fb_tests which includes the same test run here.

Can you also check the other that were added to make sure we're not running it twice?

travagliad: I think it's not duplicating anywhere else

e2e_tests/tests/serverLogs.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(response.status(), response.statusText()).toEqual(200);
+
+    const actualLineCount = getFileLineCount(await response.body(), fileNameToCheck);
+
+    assertLineCount(
+      actualLineCount,
+      `File ${fileNameToCheck} has ${actualLineCount} lines, but expected ${expectation}`,
+    );
+  });
+}
```
travagliad: I think this is super confusing for readability

inlining the tests here might not even waste this many lines, but even if we parametrize I think it should take few params and make it more simple with expects coming to here instead.

What you think?

.github/workflows/e2e-tests-matrix.yml
kiranvuyurru: Agreed and dropped in 7b0b2d7 — `e2e-tests-matrix.yml` is byte-identical with `main` again, so `@fb-settings` now runs once per PR-CI build, via `fb_tests` → the `settings` job.

Confirming your follow-up: it isn't duplicating anywhere else. This PR added exactly two jobs and they were the pair, so there's nothing left to remove. I swept both files for same-tag pairs and the only others are `docker_configuration` / `docker_configuration_playwright` and `user_and_password` / `user_and_password_podman` — both pre-existing, and both deliberately spanning *different* runners rather than running one suite twice, so I left them alone.

One note for the record: `alerting_and_settings` still greps `@fb-alerting|@fb-settings` on the CodeceptJS runner, but that is not a third run of these tests — retiring `serverLogs_test.js` off the `_test.js` discovery glob took PMM-T1902-1904 out of it, so it now matches only `configuration/permissions_test.js` and `configuration/verifyPMMSettingsPageFunctionality_test.js`.

---

e2e_tests/tests/serverLogs.test.ts
kiranvuyurru: You're right, and I went with inlining — c894169.

The table came from your first review ("`for (const c of cases) pmmTest(...)` would keep them as three separate tests without repeating the body"), but what it actually removed was ~8 lines of mechanics, and the price was an `assertLineCount` callback sitting in what should be plain data. That's the confusing part, and it isn't fixable by trimming params: two scenarios assert `toBe` and the third `toBeGreaterThan`, so any data-only table needs a comparison flag, which reads no better.

So each test now spells out its own request and expectation, which is also the shape you called right in that same review. Three lanes, as before — the loop never changed that.

Kept from the earlier round: the `expect(response.status(), response.statusText()).toEqual(200)` on all three, so a 401/500 fails on the status line instead of surfacing as `Invalid or unsupported zip format` out of `getFileLineCount`.

Verified: `npm run lint` (eslint + tsc) clean, prettier clean, and `--list --grep "@fb-settings"` still reports PMM-T1902/1903/1904 with identical titles.

---

travagliad (approved):

PR #1220, opened 2026-08-20
migrate(dashboards): verifyAnnotations CodeceptJS to Playwright

kiranvuyurru:
Migrates the annotation dashboard suite from CodeceptJS to native Playwright.

## Source

`codeceptjs-e2e/tests/verifyAnnotations_test.js`, retired here by renaming to
`verifyAnnotations_migrated.js` so it no longer matches the CodeceptJS discovery glob
`tests/**/*_test.js` (`pr.codecept.js:147`, `launchable-prepare.js:18`).

`annotationAPI.js`, its `codeceptConfigHelper.js` registration and the `dashboardPage` annotation
helpers are deliberately kept: `tests/upgrade/annotationsPrometheus_test.js` still uses them.

## Target

`e2e_tests/tests/verifyAnnotations.test.ts` (new file).

## Migrated scenarios and preserved tags

| Scenario | Tags |
| --- | --- |
| PMM-T878 - Verify adding annotation specific dashboard (data-driven, 5 rows, 1 row on the gssapi job) | `@nightly  @dashboards @annotations` |
| PMM-T878 - Verify user is not able to add an annotation for non-existing node name or service name and without service name | `@nightly  @dashboards` + `@annotations` |
| PMM-T165 - Verify Annotation with Default Options | `@fb-instances` |
| PMM-T166 - Verify adding annotation with specified tags | `@fb-instances` |

Every original CodeceptJS tag is preserved verbatim, including the double space in
`@nightly  @dashboards`. `@annotations` is added to the negative scenario as a destination
execution tag only: the nightly Playwright matrix greps per-scenario titles, so without it that
scenario would have had no Playwright home once the source retires. No source tag was replaced.

The `isJenkinsGssapiJob` row switch and the `DataTable` row set are preserved, as are the `Before`
authorization hook, all four status-code assertions (200 / 404 / 404 / 400), the per-scenario
annotation marker index, and the annotation text and tag-text assertions.

## Source and target files queried from the existing graphs

Source: `tests/verifyAnnotations_test.js`, `tests/helper/constants.js`, `tests/pages/dashboardPage.js`,
`tests/pages/api/annotationAPI.js`, `tests/pages/api/inventoryAPI.js`,
`tests/configuration/pages/pmmInventoryPage.js`, `tests/custom_steps.js`.

Target: `fixtures/pmmTest.ts`, `api/api.ts`, `api/inventory.api.ts`, `helpers/apiEndpoints.ts`,
`helpers/url.helper.ts`, `helpers/cli.helper.ts`, `helpers/timeouts.ts`, `helpers/grafana.helper.ts`,
`interfaces/inventory.ts`, `pages/dashboards/dashboards.page.ts`, `pages/dashboards/mysql/`,
`pages/dashboards/mongo/`, `pages/dashboards/operating-system/`, `pages/inventory/services.page.ts`.

Both graphs were refreshed before the migration started (`graphify . --update --code-only`, no LLM
API key available locally): Playwright 692 nodes / 1456 edges, CodeceptJS 1851 nodes / 2567 edges.
Graph artifacts are excluded from this PR.

## Reused and added Playwright code

Reused: the `pmmTest` fixture, `grafanaHelper.authorize()`, `cliHelper.execute().assertSuccess()`,
`urlHelper.buildUrlWithParameters()`, `dashboard.waitForDashboardToLoad()`,
`api.inventoryApi.getServiceDetailsByPartialName()`, `ServiceType`, `Timeouts`, and
`MysqlDashboards.mysqlInstanceSummary`.

Added: `api/annotation.api.ts` (`setAnnotation`, registered in `api/api.ts` and
`helpers/apiEndpoints.ts`), `InventoryApi.getNodeName`, `InventoryApi.getServiceDetailsByTypeAndPartialName`,
a `nodeName` parameter on the url helper, three annotation locators on the dashboards POM, and the
PostgreSQL instance summary, MongoDB instance summary, nodes compare and processes details
dashboard URLs.

## Setup used

`--database ps --database psmdb --database pgsql --database haproxy`, `setup_client=false`,
provisioned locally through `provisioning/setup.ts`.

## Static validation

`npx tsc --noEmit` clean and `npx eslint` clean on all 15 changed TypeScript files, run in this
branch's own worktree after the cherry-pick. Migration conventions check clean.

## MCP locator verification

All three new annotation locators were verified against the live DOM of the provisioned PMM
environment. Iframe scoping was proven required, plain `graph/d/...` navigation was confirmed, and
`ServiceType.mongodb = 'mongodb'` was confirmed live. Zero locator corrections were needed.

## Execution

Complete file, declaration order, one worker, against the provisioned local PMM:

```
PMM_MIGRATION=1 PMM_UI_URL=https://127.0.0.1/ ADMIN_PASSWORD=admin \
  npx playwright test tests/verifyAnnotations.test.ts --workers=1
-> 8 passed (1.6m)
```

Run from this branch's freshly cherry-picked worktree, not only from the development branch.
Target-file regression is not required: this is a new file and the whole file was run.

Note for anyone rerunning this locally: marker selection is index-based (faithful to the source's
`verifyAnnotationsLoaded(name, 1|2)`), so a reused PMM needs its Grafana annotation table emptied
between runs. CI always starts from a fresh server.

## Reviews

Initial independent review plus MCP locator verification: `READY_TO_RUN`, zero findings, zero
locator edits. Final post-run review: `FINAL_REVIEW_PASS`.

## Workflow coverage

Existing CodeceptJS jobs and grep expressions are unchanged.

- `.github/workflows/nightly-e2e-tests-matrix.yml`: adds the `test_execution_playwright` job
  greping `@annotations`, and raises the setup shards' `expected_test_jobs` from 2 to 3. That
  count is load-bearing: the shards poll for job names prefixed `test execution / ` to know when
  it is safe to stop waiting, so undercounting would let a shard terminate while this consumer is
  still running. The 14 setup shards already provide ps, psmdb, pgsql and haproxy.
- `.github/workflows/fb-e2e-suite.yml`: adds `instances_playwright`, greping `@fb-instances` with
  the same `setup_services` as the CodeceptJS `instances` job, because PMM-T165/T166 shell into
  the `haproxy_pmm` container and no existing Playwright job in that file provides haproxy. This
  follows the `docker_configuration` / `docker_configuration_playwright` pairing already there.
  The CodeceptJS `instances` job is untouched; three other CodeceptJS tests still carry
  `@fb-instances`.
- `.github/workflows/e2e-tests-matrix.yml`: no change. It already reaches the new FB job through
  its `fb_tests` job, and it carries no `@nightly` coverage.

Both tags were verified to select the intended scenarios and nothing else, with
`npx playwright test --list --grep` (`@annotations` -> the 6 PMM-T878 scenarios, `@fb-instances`
-> exactly PMM-T165 then PMM-T166, union = the full file) and with `launchable-prepare.js`, which
returns a non-empty subset for both.


kiranvuyurru:
GitHub Actions: https://github.com/percona/pmm-qa/actions/runs/32429363442

kiranvuyurru:
@claude review

.github/workflows/nightly-e2e-tests-matrix.yml
```diff
@@ -170,6 +170,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - tags_for_tests: '@annotations'
```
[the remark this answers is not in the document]
kiranvuyurru: Confirmed and fixed in 3e8909f. `expected_test_jobs` is now 4 (2 CodeceptJS `test_execution` entries + 2 `test_execution_playwright` entries), and the comment above it is updated to name both Playwright matrix jobs plus a reminder to bump it whenever either matrix gains an entry.

---

e2e_tests/api/inventory.api.ts
```diff
@@ -48,6 +60,25 @@ export default class InventoryApi {
     return filteredServices[0];
   };
 
+  getServiceDetailsByTypeAndPartialName = async (
+    serviceType: ServiceType,
+    partialServiceName: string,
+    excludeSubstring?: string,
```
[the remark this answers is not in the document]
kiranvuyurru: Confirmed and fixed in 3e8909f. There is no such collision in any environment, so the parameter is gone: `getServiceDetailsByTypeAndPartialName` now takes two arguments and resolves `services.at(0)`, and the `'pmm-server'` argument is dropped from the call site. The other call site (`verifyAnnotations.test.ts:122`) never passed it. `npm run lint` (eslint + tsc) is clean.

---

e2e_tests/pages/dashboards/postgresql/index.ts
```diff
@@ -0,0 +1,10 @@
+import DashboardInterface from '@interfaces/dashboard';
+import PostgresqlInstanceSummary from './postgresqlInstanceSummary';
+
+export const PostgresqlDashboards = {
+  postgresqlInstanceSummary: new PostgresqlInstanceSummary(),
+};
+
+export type PostgresqlDashboardsType = typeof PostgresqlDashboards & Record<string, DashboardInterface>;
```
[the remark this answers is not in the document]
kiranvuyurru: Wired it up in 3e8909f — `readonly postgresql: PostgresqlDashboardsType = PostgresqlDashboards;` now sits alongside `mongo`/`mysql`/`os`/`valkey` on `Dashboards`, so the type has a consumer and `dashboard.postgresql` exists like its siblings.

On the 🔵 naming point: withdrawing it. `MysqlDashboards` — the largest registry — uses the prefixed form throughout (`mysqlInstanceSummary`, `mysqlUserDetails`, …), so the repo already carries both conventions and `postgresqlInstanceSummary` matches one of them rather than inventing a third. Renaming it would only trade which sibling it disagrees with, so I left it as-is.

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -31,6 +31,10 @@ export default class Dashboards extends BasePage {
   readonly os: OperatingSystemDashboardsType = OperatingSystemDashboards;
   readonly valkey: ValkeyDashboardsType = ValkeyDashboards;
   builders = {
+    annotationTagText: (tagValue: string) =>
+      this.grafanaIframe().locator(`//span[contains(text(), '${tagValue}')]`),
+    annotationText: (annotationTitle: string) =>
+      this.grafanaIframe().locator(`//div[contains(text(), '${annotationTitle}')]`),
```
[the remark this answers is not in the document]
kiranvuyurru: Not taking this one. `getByText` would widen the match rather than narrow it: the XPath is element-scoped (`//div[...]` for the title, `//span[...]` for the tag), whereas `getByText` matches whatever element carries the text. Both builders are asserted with `toBeVisible()`, so any second match anywhere in the dashboard iframe — an annotation title that also appears in a panel legend or axis label, for instance — turns a passing assertion into a strict-mode violation. That is a behaviour change I cannot validate here (these tests need a live PMM server with annotations present), and it would be traded for style alone.

The `loadingText` comparison doesn't carry over either: that one is a fixed string with `{ exact: true }`, not a substring match on caller-supplied text.

Leaving the XPath, consistent with the pre-existing XPath uses in this file.

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,202 @@
...
+pmmTest.beforeEach(async ({ grafanaHelper }) => {
+  await grafanaHelper.authorize();
+});
+
+for (const annotation of annotations) {
+  pmmTest(
+    `PMM-T878 - Verify adding annotation specific dashboard @nightly  @dashboards @annotations | ${JSON.stringify(annotation)}`,
+    async ({ api, dashboard, page, servicesPage, urlHelper }) => {
+      const isNodeAnnotation = annotation.annotationName === nodeAnnotationName;
+      const annotationMarker = dashboard.elements.annotationMarkers.nth(isNodeAnnotation ? 1 : 0);
```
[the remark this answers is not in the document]
kiranvuyurru: Decision made: matching markers by content. Fixed in 142ba2c.

New `Dashboards.hoverAnnotationMarker(annotationTitle)` hovers each marker in turn until one shows the expected title, and all three annotation tests use it — `nth()`/`first()` and the fixed indices are gone (`verifyAnnotations.test.ts:64`, `162-164`, `183-189`). That makes the suite retry- and worker-safe without a cleanup hook: PMM exposes no annotation-delete endpoint, and content matching doesn't need one, since a duplicate marker from a retry carries the same title and either one satisfies the assertion.

One thing worth noting from doing it: content matching *introduced* a collision the positional lookup had accidentally avoided. `annotationText` is `//div[contains(text(), …)]`, so a scan for `annotation-for-postgres` would have happily accepted the tooltip of `annotation-for-postgres-server` — both live on the same PostgreSQL Instance Summary dashboard. The helper therefore compares tooltip text exactly (`text.trim() === annotationTitle`) using the proven `contains()` locator only to find candidates, rather than tightening the shared builder's XPath to an exact match I can't validate without a live server.

Unique titles were considered and skipped: they'd make a duplicate identifiable but wouldn't fix the index shift on their own, which was the actual defect.

`npm run lint` (eslint + tsc) is clean.

---

kiranvuyurru:
## `@fb-instances` failure on 142ba2c — root-caused on a live PMM box

142ba2c (marker matching by tooltip content) broke `@fb-instances`. `@fb-instances` was green on every earlier commit through 3e8909f, so this was my regression. Reproduced and fixed on a throwaway Linode PMM box (`ps` + `external` + `haproxy` + `psmdb` + `pgsql`, `perconalab/pmm-server:3-dev-latest`). Two independent bugs were stacked; the CI log only exposed the first.

**1. Unhoverable markers aborted the scan** — fixed in 48834e5

Processes Details renders **24** annotation markers for two annotations: every panel draws its own marker per annotation, and annotations minutes apart land within a few pixels of each other (`left: 565px` beside `left: 571px`). Measured on the box, **9 of 24** have another marker over their centre, so Playwright refuses to hover them:

```
<div class="css-17w0old" data-testid="data-testid annotation-marker"> intercepts pointer events
```

The scan called `hover()` unguarded, so the first such marker threw at the project's 10s `actionTimeout` and ended the search. The positional version never hit this because it hovered exactly one marker and never iterated. Fix: hover inside the try/catch with a short timeout and move to the next marker.

**2. The tooltip title check could never match an API annotation** — fixed in 11cfe71

This is the one that mattered more, and it was invisible to `@fb-instances`. An API-created annotation does not render its title alone. Captured live:

```
annotation-for-mysql (Service Name: ps_pmm_8_0_1_29315. Node Name: 8b9a9a999d74)
```

so `text.trim() === annotationTitle` never matched. `pmm-admin annotate` renders the bare title, which is why PMM-T165/T166 passed while **every** PMM-T878 case failed. The check now accepts the bare title or the title followed by `" ("`, which still keeps `annotation-for-postgres` from accepting `annotation-for-postgres-server`'s tooltip.

### Verified on the box (`--grep "@annotations|@fb-instances"`)

| Version | Result |
|---|---|
| 142ba2c (as pushed) | `@fb-instances` **2 failed** |
| 3e8909f (positional — the faithful port) | `@annotations` **3 passed / 3 failed** |
| 11cfe71 (this head) | **8 passed** (2.2m) — all 6 `@annotations` + both `@fb-instances` |

Worth flagging: the positional version — the shape the CodeceptJS source used — **also failed 3 of 6** PMM-T878 cases in that environment (`annotation-for-postgres-server`, `annotation-for-mongo`, `annotation-for-postgres`). Those are `@nightly`-only, so no PR CI ever exercised them and the breakage was never visible. Content matching plus the two fixes is the first version to pass all six, so this ends up ahead of the original port rather than merely level with it.

### Functionality checked against the CodeceptJS source

Compared `verifyAnnotations_migrated.js` and `dashboardPage.verifyAnnotationsLoaded` line by line; nothing else diverges:

- `annotationLocator(1)`/`(2)` (1-indexed) mapped correctly onto `nth(0)`/`nth(1)`, and both are now replaced by content matching.
- URL params match: CodeceptJS `node_name`/`service_name` and the Playwright helper's `nodeName`/`serviceName` both emit `var-node_name`/`var-service_name`; CodeceptJS defaults `from=now-5m&to=now`, and the port passes `to: 'now'` explicitly.
- `.url` vs `.clearUrl`/`.cleanUrl`: the CodeceptJS `url` variants that carry a query string (`processDetailsDashboard.url` = `…?from=now-45m&to=now`) are reproduced by the port's bare `url` plus `buildUrlWithParameters`.
- The two-step navigation (bare dashboard, then filtered) is preserved.
- `annotationTagText`/`annotationText` locators are unchanged from the source.

The `getServiceDetailsByTypeAndPartialName` change from 3e8909f is confirmed sound against real inventory: `getServicesByType(postgresql)` drops `pmm-server-postgresql` before the partial-name filter, so `pgsql` resolved to `pgsql_pgss_pmm_17_service_6112` and `pmm-server` to `pmm-server-postgresql` via `getServiceDetailsByPartialName` — the removed `excludeSubstring` had nothing to exclude.

VM destroyed. `npm run lint` (eslint + tsc) clean; Lint green on this head; `@fb-instances` re-running now.

---

kiranvuyurru:
Two follow-ups on the latest review.

### XPath escaping (`dashboards.page.ts` 36-39) — valid, fixed in c56ee5e

Confirmed in Chromium: the old form throws before it can assert anything.

```
SyntaxError: Failed to execute 'evaluate' on 'Document':
The string '//div[contains(text(), 'customer's deploy')]' is not a valid XPath expression.
```

XPath 1.0 string literals have no escape syntax and cannot mix quote styles, so the value is now assembled with `concat()` when it contains an apostrophe (new `@helpers/xpath.helper`). Verified across bare, embedded, leading and trailing apostrophes plus the API tooltip form — and a value **without** an apostrophe yields the byte-identical selector as before, so the locator behaviour measured on the live box is unchanged.

Scoped to the two annotation builders this PR adds. `panelByName` and `noDataMarkerXPath` interpolate too, but they are pre-existing and take panel/metric names rather than user text, so widening the PR to them isn't warranted here.

### Correcting the learning recorded on the resolved thread

The learning captured from [#discussion_r3925265613](https://github.com/percona/pmm-qa/pull/1220#discussion_r3925265613) says the check "must use `text.trim() === annotationTitle`". That was my claim, and **live measurement disproved it** — please don't reapply it. An API-created annotation does not render its title alone:

```
annotation-for-mysql (Service Name: ps_pmm_8_0_1_29315. Node Name: 8b9a9a999d74)
```

so exact equality never matched and **every** PMM-T878 case failed. `pmm-admin annotate` renders the bare title, which is why PMM-T165/T166 still passed and the breakage stayed hidden. 11cfe71 replaced it with a boundary match — the bare title, or the title followed by `" ("` — which still keeps `annotation-for-postgres` from accepting `annotation-for-postgres-server`'s tooltip. That distinction was the whole point of the original exact check, and it survives.

The corrected rule, if it's worth recording: **match the annotation title at a boundary, not as a whole string**, because the tooltip carries a `(Service Name: …. Node Name: …)` suffix for API annotations and no suffix for CLI ones.

Verification for all of the above is in the [previous comment](https://github.com/percona/pmm-qa/pull/1220#issuecomment-5528875981): 8 passed on a live PMM box, versus 3 of 6 `@annotations` for the positional shape the CodeceptJS source used.

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -97,6 +105,55 @@ export default class Dashboards extends BasePage {
...
+    // Markers carry no attribute naming their annotation, so identifying one
+    // means hovering it and reading the tooltip. Picking a marker by position
+    // instead breaks under `retries`: a retry re-runs the annotation setup, and
+    // the extra marker shifts every index after it.
+    const tooltip = this.builders.annotationText(annotationTitle);
+    // `annotationText` matches on a substring, so "annotation-for-postgres"
+    // would otherwise accept the tooltip of "annotation-for-postgres-server".
+    // Match the title at a boundary instead of comparing whole strings: a
+    // `pmm-admin annotate` tooltip is the bare title, while an API annotation
+    // renders as `<title> (Service Name: x. Node Name: y)`.
```
travagliad: trim comments

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -97,6 +105,55 @@ export default class Dashboards extends BasePage {
     return Array.from(collected);
   };
 
+  hoverAnnotationMarker = async (annotationTitle: string, timeout: Timeouts = Timeouts.TWO_MINUTES) => {
+    // Markers carry no attribute naming their annotation, so identifying one
+    // means hovering it and reading the tooltip. Picking a marker by position
+    // instead breaks under `retries`: a retry re-runs the annotation setup, and
+    // the extra marker shifts every index after it.
```
travagliad: trim comments

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -97,6 +105,55 @@ export default class Dashboards extends BasePage {
...
+
+          try {
+            // Every panel draws its own marker for each annotation, and two
+            // annotations minutes apart land within a few pixels of each other,
+            // so a marker whose centre another marker covers cannot be hovered
+            // at all -- Playwright reports "intercepts pointer events". That is
+            // ordinary here (9 of 24 markers on Processes Details), not a
+            // failure: skip to the next marker instead of letting it end the
+            // search. The short timeout keeps a full sweep well inside the
+            // budget, since the project's actionTimeout is 10s.
```
travagliad: trim comments

.github/workflows/nightly-e2e-tests-matrix.yml
```diff
@@ -123,12 +123,13 @@ jobs:
...
-      # job-name prefix "test execution / " to know when it's safe to finish
-      # waiting -- undercounting here lets a shard finish (and its runner
-      # terminate) while a still-running consumer job needs it.
-      expected_test_jobs: 3
+      # 4: the two CodeceptJS `test_execution / ...` jobs plus the two
+      # Playwright `test_execution / playwright ...` matrix jobs below. The
+      # setup shards poll by job-name prefix "test execution / " to know when
+      # it's safe to finish waiting -- undercounting here lets a shard finish
+      # (and its runner terminate) while a still-running consumer job needs
+      # it, so bump this whenever either matrix gains an entry.
```
travagliad: trim comments (AI is going crazy here, adding more lines each time it touches it)

kiranvuyurru: Fair, and accurate — the comment grew on every touch. Trimmed in 15f4478.

- `expected_test_jobs`: back to `main`'s wording verbatim. My collapsing of the Playwright matrix into one job restored the count to 3, so there was nothing left to explain and the diff here is now zero.
- The block above `test_execution_playwright`: 14 lines → 5, one fewer than `main`, and the reference to `.claude/skills/codeceptjs-migration/branch-workflow.md` is gone — that file does not exist in the repo.

Net across this file: +5 / −6 comment lines against `main`.

---

e2e_tests/pages/dashboards/dashboards.page.ts
kiranvuyurru: Trimmed in 15f4478 — `hoverAnnotationMarker` goes from 16 comment lines to 3 one-liners, which matches this file's density. The three findings the prose was recording are in the commit messages and the PR comments instead.

---

e2e_tests/pages/dashboards/dashboards.page.ts
kiranvuyurru: Trimmed in 15f4478 — the tooltip-matching explanation is now one line: `// An API annotation's tooltip appends " (Service Name: x. Node Name: y)".`

---

e2e_tests/pages/dashboards/dashboards.page.ts
kiranvuyurru: Trimmed in 15f4478 — the 8-line note inside the loop is now one line: `// Overlapping markers cover each other, so some cannot be hovered.`

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -29,8 +31,13 @@ export default class Dashboards extends BasePage {
   readonly mongo: MongoDashboardsType = MongoDashboards;
   readonly mysql: MysqlDashboardsType = MysqlDashboards;
   readonly os: OperatingSystemDashboardsType = OperatingSystemDashboards;
+  readonly postgresql: PostgresqlDashboardsType = PostgresqlDashboards;
   readonly valkey: ValkeyDashboardsType = ValkeyDashboards;
   builders = {
+    annotationTagText: (tagValue: string) =>
+      this.grafanaIframe().locator(`//span[contains(text(), ${xpathLiteral(tagValue)})]`),
+    annotationText: (annotationTitle: string) =>
+      this.grafanaIframe().locator(`//div[contains(text(), ${xpathLiteral(annotationTitle)})]`),
```
travagliad: 🟡 `xpathLiteral` guards an input this suite never produces (every title and tag is a constant in the test file), and it is only needed because the text is spliced into selector source. Let Playwright quote it instead: in the Grafana PMM ships, the open tooltip has its own test id (`AnnotationMarker2.tsx:104` on percona/grafana `main`: `data-testid="annotation-marker"`, distinct from the marker's `data-testid annotation-marker`; `getByTestId` matches exactly). Scope to it and use `getByText`, which takes the value as data and returns the innermost element, i.e. the same `span`/`div` these XPaths target, limited to the open tooltip instead of the whole iframe. Together with the companion suggestions below, `annotationText`, `tooltipShowsTitle` and `e2e_tests/helpers/xpath.helper.ts` all go away.

```suggestion
    annotationTagText: (tagValue: string) =>
      this.elements.annotationTooltip.getByText(tagValue, { exact: true }),
```

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -3,10 +3,12 @@ import { GrafanaPanel } from '@interfaces/grafanaPanel';
 import { GetService } from '@interfaces/inventory';
 import { replaceWildcards } from '@helpers/metrics.helper';
 import { Timeouts } from '@helpers/timeouts';
+import { xpathLiteral } from '@helpers/xpath.helper';
```
travagliad: Companion to the `annotationTagText` thread: nothing imports `xpathLiteral` once the builders use `getByText`; delete `e2e_tests/helpers/xpath.helper.ts` with it.

```suggestion
```

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -46,6 +53,7 @@ export default class Dashboards extends BasePage {
     imageRendererGenerateImage: this.grafanaIframe().getByRole('button', { name: 'Generate image' }),
   };
   elements = {
+    annotationMarkers: this.grafanaIframe().getByTestId('data-testid annotation-marker'),
```
travagliad, line 36: Companion to the `annotationTagText` thread: the tooltip locator the builders and `hoverAnnotationMarker` scope to.

```suggestion
    annotationMarkers: this.grafanaIframe().getByTestId('data-testid annotation-marker'),
    annotationTooltip: this.grafanaIframe().getByTestId('annotation-marker'),
```

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -97,6 +105,41 @@ export default class Dashboards extends BasePage {
...
 
+  hoverAnnotationMarker = async (annotationTitle: string, timeout: Timeouts = Timeouts.TWO_MINUTES) => {
+    // Markers are unlabelled, so the tooltip is the only way to tell them apart.
+    const tooltip = this.builders.annotationText(annotationTitle);
+    // An API annotation's tooltip appends " (Service Name: x. Node Name: y)".
+    const tooltipShowsTitle = async () =>
+      (await tooltip.allTextContents())
+        .map((text) => text.trim())
+        .some((text) => text === annotationTitle || text.startsWith(`${annotationTitle} (`));
+    const deadline = Date.now() + timeout;
```
travagliad, line 113: Companion to the `annotationTagText` thread: the boundary check becomes a locator, two plain strings and no escaping. `exact: true` matches the bare title `pmm-admin annotate` renders; the second alternative matches the API form and still rejects `annotation-for-postgres-server` for `annotation-for-postgres`.

```suggestion
    // Markers are unlabelled, so the tooltip is the only way to tell them apart.
    // API annotations render `title (Service Name: x. Node Name: y)`, CLI ones the bare title.
    const tooltipTitle = this.elements.annotationTooltip
      .getByText(annotationTitle, { exact: true })
      .or(this.elements.annotationTooltip.getByText(`${annotationTitle} (`));
    const deadline = Date.now() + timeout;
```

---

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -97,6 +105,41 @@ export default class Dashboards extends BasePage {
...
+
+          try {
+            // Overlapping markers cover each other, so some cannot be hovered.
+            await this.elements.annotationMarkers.nth(i).hover({ timeout: Timeouts.THREE_SECONDS });
+            await tooltip.first().waitFor({ state: 'visible', timeout: Timeouts.TWO_SECONDS });
+          } catch {
+            continue;
+          }
+
+          if (await tooltipShowsTitle()) return;
```
travagliad: Companion to the `annotationTagText` thread: with the title check inside the locator, the wait is the check.

```suggestion
          try {
            // Overlapping markers cover each other, so some cannot be hovered.
            await this.elements.annotationMarkers.nth(i).hover({ timeout: Timeouts.THREE_SECONDS });
            await tooltipTitle.waitFor({ state: 'visible', timeout: Timeouts.TWO_SECONDS });
            return;
          } catch {
            continue;
          }
```

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        `Add annotation ${annotation.annotationName} for ${annotation.service}`,
+        async () => {
+          const service =
+            annotation.service === 'pmm-server'
+              ? await api.inventoryApi.getServiceDetailsByPartialName(annotation.service)
+              : await api.inventoryApi.getServiceDetailsByTypeAndPartialName(
+                  annotation.service_type,
+                  annotation.service,
+                );
+          const node = await api.inventoryApi.getNodeName(service.node_id);
```
travagliad: 🟡 `GetService` already carries `node_name` (`e2e_tests/interfaces/inventory.ts:27`, from `UniversalService.node_name` in the management API), so this is a second HTTP round trip per row for a field the previous call returned. With it, `InventoryApi.getNodeName`, the `GetNode` interface and `apiEndpoints.management.nodes` have no caller left; suggestions on those files follow.

```suggestion
          const node = service.node_name;
```

---

e2e_tests/api/inventory.api.ts
```diff
@@ -1,11 +1,23 @@
...
+      headers: GrafanaHelper.getAuthHeader(),
+    });
+    const nodes = Object.values((await response.json()) as Record<string, GetNode>).flat();
+    const node = nodes.find((node: GetNode) => node.node_id === nodeId);
+
+    if (!node) throw new Error(`Node with id ${nodeId} is not present`);
+
+    return node.node_name;
+  };
+
```
travagliad: Companion to the `getNodeName` thread on the test: no caller once the test reads `service.node_name`.

```suggestion
```

---

e2e_tests/api/inventory.api.ts
```diff
@@ -1,11 +1,23 @@
 import { APIRequestContext, expect } from '@playwright/test';
 import GrafanaHelper from '@helpers/grafana.helper';
-import { AgentStatus, GetService, GetServices, ServiceType } from '@interfaces/inventory';
+import { AgentStatus, GetNode, GetService, GetServices, ServiceType } from '@interfaces/inventory';
```
travagliad: Companion to the `getNodeName` thread on the test.

```suggestion
import { AgentStatus, GetService, GetServices, ServiceType } from '@interfaces/inventory';
```

---

e2e_tests/interfaces/inventory.ts
```diff
@@ -1,8 +1,14 @@
+export interface GetNode {
+  node_id: string;
+  node_name: string;
+}
+
```
travagliad: Companion to the `getNodeName` thread on the test: `GetNode` has no consumer once `getNodeName` goes.

```suggestion
```

---

e2e_tests/helpers/apiEndpoints.ts
```diff
@@ -21,6 +21,8 @@ const apiEndpoints = {
     services: '/v1/inventory/services',
   },
   management: {
+    annotations: '/v1/management/annotations',
+    nodes: '/v1/management/nodes',
```
travagliad: Companion to the `getNodeName` thread on the test: unused once `getNodeName` goes.

```suggestion
```

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        service: 'ps_',
+        service_type: ServiceType.mysql,
+      },
+      {
+        annotationName: nodeAnnotationName,
+        dashboard: OperatingSystemDashboards.nodesCompare.url,
+        service: 'ps_',
+        service_type: ServiceType.mysql,
+      },
+    ];
```
travagliad: 🟡 This branch is unreachable: `JOB_NAME` is a Jenkins variable, and the only runner of `@annotations` is the GitHub Actions nightly matrix, which never sets it. The Jenkins gssapi job (`pmm3-ui-tests-nightly-gssapi.groovy`) runs CodeceptJS with `--grep '@gssapi-nightly'`, a tag this suite never carried, so dropping the branch loses no coverage; if gssapi coverage is wanted it needs a real input (a workflow env or tag), not a dead constant.

```suggestion
const nodeAnnotationName = 'mysql-node-name';
const processesDetailsUrl = OperatingSystemDashboards.processesDetails.url;
const annotations = [
  {
    annotationName: 'annotation-for-postgres-server',
    dashboard: PostgresqlDashboards.postgresqlInstanceSummary.url,
    service: 'pmm-server',
    service_type: ServiceType.postgresql,
  },
  {
    annotationName: 'annotation-for-mongo',
    dashboard: MongoDashboards.instanceSummary.url,
    service: 'rs101',
    service_type: ServiceType.mongodb,
  },
  {
    annotationName: 'annotation-for-postgres',
    dashboard: PostgresqlDashboards.postgresqlInstanceSummary.url,
    service: 'pgsql',
    service_type: ServiceType.postgresql,
  },
  {
    annotationName: 'annotation-for-mysql',
    dashboard: MysqlDashboards.mysqlInstanceSummary.url,
    service: 'ps_',
    service_type: ServiceType.mysql,
  },
  {
    annotationName: nodeAnnotationName,
    dashboard: OperatingSystemDashboards.nodesCompare.url,
    service: 'ps_',
    service_type: ServiceType.mysql,
  },
];
```

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+});
+
+for (const annotation of annotations) {
+  pmmTest(
+    `PMM-T878 - Verify adding annotation specific dashboard @nightly  @dashboards @annotations | ${JSON.stringify(annotation)}`,
+    async ({ api, dashboard, page, servicesPage, urlHelper }) => {
+      const isNodeAnnotation = annotation.annotationName === nodeAnnotationName;
+
+      await page.goto(servicesPage.url);
+
```
travagliad: 🟡 This navigation asserts nothing and nothing after it needs a page: `authorize()` in `beforeEach` sets the header and the session cookie on the context, the inventory and annotation calls go through `request`, and the dashboard `goto` below is the first thing that needs the browser. It only exists because the source did `I.amOnPage(pmmInventoryPage.url)`, and it drags the `servicesPage` fixture in for a `.url` string. Same at line 114.

```suggestion
    async ({ api, dashboard, page, urlHelper }) => {
      const isNodeAnnotation = annotation.annotationName === nodeAnnotationName;

```

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      });
+    },
+  );
+}
+
+pmmTest(
+  'PMM-T878 - Verify user is not able to add an annotation for non-existing node name or service name and without service name @nightly  @dashboards @annotations',
+  async ({ api, page, servicesPage }) => {
+    await page.goto(servicesPage.url);
+
```
travagliad: Companion to the `servicesPage` thread above: this test is API-only.

```suggestion
  async ({ api }) => {
```

---

e2e_tests/tests/verifyAnnotations.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      await pmmTest.step(`Verify ${annotation.annotationName} is loaded on the dashboard`, async () => {
+        const filteredDashboardUrl = urlHelper.buildUrlWithParameters(
+          annotation.dashboard,
+          isNodeAnnotation
+            ? { from: 'now-1h', nodeName, refresh: '5s', to: 'now' }
+            : { from: 'now-1h', refresh: '5s', serviceName, to: 'now' },
+        );
+
+        await page.goto(annotation.dashboard);
+        await dashboard.waitForDashboardToLoad();
```
travagliad: 🟡 Loading the unfiltered dashboard first verifies nothing and is discarded by the filtered `goto` on the next line: a full dashboard load per row, five rows. Every other dashboard test in this suite goes straight to the filtered URL (`mysqlDashboards.test.ts:32`), and the migration is the moment to drop a step that was already inert in the source.

```suggestion
```

---

travagliad (approved):

PR #1246, opened 2026-08-25
PMM-15302 Verify no SSH key settings off AMI

travagliad:
## What

One Playwright test for the half of [percona/pmm#5744](https://github.com/percona/pmm/pull/5744) that had no coverage anywhere: on a deployment that is **not** AMI, the Settings page must not offer an SSH key tab.

**PMM-T2282** — on the suite's own Docker server: no `settings-tab-ssh` in the DOM, `/pmm-ui/settings/ssh-key` redirects to the default tab, and `PUT /v1/server/settings` with an `ssh_key` comes back `500 … SSH key can be set only on AMI distribution` with the stored setting unchanged.

Tagged `@docker-configuration`, so it rides the existing Playwright job in `fb-e2e-suite.yml`. `e2e-tests-matrix.yml` calls that suite as its `fb_tests` job, so it runs on every pmm-qa PR and on that workflow's 02:00 cron.

## The AMI side is deliberately not here

An earlier revision had a second test, `PMM-T2283`, that started a container with `PMM_DISTRIBUTION_METHOD=ami`. That makes the server *report* AMI — which is what the gate keys off — but it is not an AMI deployment: no EC2, no AMI image, no `admin` OS user, and SSH never attempted. Simulating AMI to assert AMI behaviour is not worth having, so it is gone (`021668d`), and its Zephyr case is deprecated.

The AMI side belongs to the tests that run against real AMI deployments — `pmm3-ui-tests-nightly-gha` with `SERVER_TYPE=ami` (nightly cron, and via `pmm3-rc-testing`'s `nightly (AMI)` stage), which boots a real AMI instance before dispatching the suite. Worth noting for whoever owns those, since only a real AMI can catch it: `writeSSHKey` writes the key to the **pmm** OS user's home, `/home/pmm/.ssh/authorized_keys` — confirmed on a live server during manual QA — while docs PR percona/pmm#5765 tells users to `ssh admin@…`. If the AMI login account is `admin`, the key lands in a home nobody logs in as and SSH access silently never works. The username predates this change.

## Why here, and not an existing test

Checked before adding one. Nothing asserted the absence: the only other references to the SSH tab are locators plus CodeceptJS `PMM-T85`, which navigates to the SSH URL *only* when the distribution is AMI, so it never executes its assertions in CI and never covered the negative case.

`@settings` was the obvious host but is not an option: the Playwright tags any workflow selects are `@LBAC`, `@alerting`, `@docker-configuration`, `@inventory|@image-renderer`, `@new-navigation|@menu`, `@pmm-ps-integration`, `@pmm-ps-pxc-haproxy-integration`, `@pmm-psmdb-integration`, `@pmm-valkey-integration` and `@rta`. `@settings|@cli` runs on the CodeceptJS runner only, so a Playwright test tagged `@settings` — `portalRemoval.test.ts` and its siblings — does not execute. `@docker-configuration` is the Playwright job that runs on a plain Docker server, which is exactly this test's precondition.

## Supporting changes

- `ServerApi.getDistributionMethod()` — reads `distribution_method` off `/v1/version`, typed to the methods PMM ships (`AMI`, `DOCKER`, `UNSPECIFIED`).
- `SettingsResponse.ssh_key` on the settings API interface, so the refusal check can compare the stored value before and after.
- `minPmmVersion` entry for `PMM-T2282` at `3.10.0`, so it does not assert the absence of a tab that legitimately exists on a released image.

No change to any existing test, page object, or workflow.

## Validation

Green on the final head `acdcc25`: [E2E matrix #1806](https://github.com/percona/pmm-qa/actions/runs/33748339361) and `Lint` #207 both pass, all 37 check runs `success` or `skipped`. The Playwright `@docker-configuration` leg ran `PMM-T2282` for real — `✓ 5 … PMM-T2282 - Verify the SSH key tab is hidden and its URL redirects on a non-AMI deployment (2.8s)`, 7 found / 7 passed, launchable confidence 100%, 0 actionable failures — against `perconalab/pmm-server:3-dev-latest` carrying #5744. Naming the per-test line rather than the run conclusion matters here: an earlier revision was green at run level while the removed `PMM-T2283` was dying in fixture setup.

Earlier manual QA on a throwaway Linode VM also showed the test **failing** when pointed at a server reporting AMI, so the assertion has teeth. `npm run lint` (eslint + tsc) clean.

https://claude.ai/code/session_013d6xhz3WvkN1yxPy2zZCvw


travagliad:
@claude please review this.

travagliad:
@claude great review there buddy, implement it all!

travagliad:
**Hold lifted — the PR is green and ready for review.**

`percona/pmm#5744` merged to pmm `main` today at 09:37 UTC, which was the only thing this PR was waiting on. `pmm3-server-autobuild` #2699 started 11:36 UTC and published the `3-dev-latest` that carries the fix (`sha256:089421150e…`, the digest the job pulled and Docker Hub still serves for the tag).

I re-ran the one failed job on the unchanged head `2f26775`. [Job 100349511933](https://github.com/percona/pmm-qa/actions/runs/33159020010/job/100349511933):

```
✓ 5 … PMM-T2282 - Verify the SSH key tab is hidden and its URL redirects on a non-AMI deployment (4.2s)
✓ 6 … PMM-T2283 - Verify the SSH key tab is available on an AMI deployment (28.3s)
8 passed (7.3m)
```

Launchable confidence was 100%, so nothing was subset out and neither test was version-gated away — this is the first time both assertions have actually executed and passed in CI. `PMM-T2282` is the one that could not pass before #5744; it now does, against `dev-latest`, which is also the evidence that the shared `@docker-configuration` job will stay green after merge.

Changes made in response: the `DO NOT MERGE` block is out of the description, the `blocked` label is removed, and the Validation section now cites this run. No code changed, so no push — the head is the same commit the reviewers already have. All seven earlier review threads are resolved and `claude-review` is green.

@peterSirotnak @yurkovychv this is unblocked whenever you have a moment. One open question from the description worth your call rather than mine: `PMM-T85` in the CodeceptJS suite covers the same intent as `PMM-T2283` but can never execute, since no job there runs an AMI-stamped server — happy to retire its dead branch in a follow-up if you'd prefer `PMM-T2283` own that side outright.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    async ({ api, page, settingsPage }) => {
+      await pmmTest.step('Distribution method is reported as Docker', async () => {
+        expect(await api.serverApi.getDistributionMethod()).toEqual('DISTRIBUTION_METHOD_DOCKER');
+      });
+
+      await pmmTest.step('Settings page offers no SSH key tab', async () => {
+        await page.goto(settingsPage.url);
+
+        await expect(settingsPage.tabs.metrics).toHaveAttribute('aria-selected', 'true');
+        await expect(settingsPage.tabs.ssh).toHaveCount(0);
```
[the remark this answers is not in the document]
travagliad: placeholder

---

e2e_tests/helpers/versionGates.ts
```diff
@@ -8,4 +8,6 @@ export const minPmmVersion: Record<string, string> = {
   'PMM-T2266': '3.10.0',
   'PMM-T2267': '3.10.0',
   'PMM-T2268': '3.10.0',
+  'PMM-T2282': '3.10.0',
+  'PMM-T2283': '3.10.0',
```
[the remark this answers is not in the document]
travagliad: Confirmed, fixed in `5ccdf13`, and you were right that my green run proved nothing here.

I reproduced it before changing anything, running `--grep PMM-T2283` on its own so nothing could prime the cache. Identical failure point to the one you named:

```
Error: apiRequestContext.get: read ECONNRESET
  - → GET https://127.0.0.1:450/v1/version
    at ServerApi.getPmmVersion (api/server.api.ts:40:41)
    at getServerVersion (helpers/version.helper.ts:9:31)
    at Object._test.test.extend.versionGate.auto (fixtures/pmmTest.ts:182:47)
```

(`ECONNRESET` rather than `ECONNREFUSED` because the sandbox's egress proxy resets instead of refusing; same call, same stack, still in fixture setup.)

Your scheduling-order point is what I got wrong. I re-ran job [100349511933](https://github.com/percona/pmm-qa/actions/runs/33159020010/job/100349511933) after percona/pmm#5744 landed, saw `8 passed` including PMM-T2283, and reported both assertions as executing. What actually happened is that launchable ordered `clickHouse` and `clickhouseConfig` ahead of it, so `getServerVersion`'s per-worker promise was already resolved against the default `baseURL` and the `:450` request never happened. The Aug-28 run is the counter-example I should have noticed: 12 results, 6 passed, **2** actionable failures — PMM-T2282 (pre-#5744) *and* PMM-T2283. I have corrected that claim in the description and in my comment above.

Took your first option — dropped the entry rather than adding a `beforeAll`. Beyond being the smaller change, PMM-T2283 needs no gate at all: it asserts the tab is offered on an AMI-stamped server, which was true before #5744 too, since the tab then showed on every distribution. Only PMM-T2282, which asserts absence, is version-sensitive; its describe uses the default `baseURL`, so its gate resolves normally.

After the fix, same isolated run no longer touches `/v1/version` — no `version.helper.ts` in the stack — and execution reaches the test body, failing only at the `docker run` because this sandbox has no Docker daemon:

```
at ExecReturn.assertSuccess (interfaces/execReturn.ts:19:78)
at tests/dockerConfiguration/distributionMethodSshKey.test.ts:81:12
```

`PMM-T2282` still goes through `versionGate` (`pmmTest.ts:182`), so the gate I kept is intact. `npm run lint` (eslint + tsc) clean. I also left a comment on that describe recording why there is no entry, so it does not get added back.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    await page.goto('');
+    await grafanaHelper.authorize();
+  });
+
+  pmmTest(
+    'PMM-T2282 - Verify the SSH key tab is hidden and its URL redirects on a non-AMI deployment @docker-configuration',
+    async ({ api, page, settingsPage }) => {
+      await pmmTest.step('Distribution method is reported as Docker', async () => {
+        expect(await api.serverApi.getDistributionMethod()).toEqual('DISTRIBUTION_METHOD_DOCKER');
+      });
```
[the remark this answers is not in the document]
travagliad: Done in `5ccdf13` — wrapper dropped, assertion sits at the top of the test as suggested. The other three steps keep their wrappers, since each groups a navigation or a request with its assertions.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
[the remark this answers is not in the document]
travagliad: Please ignore my previous reply on this thread — I posted the word "placeholder" by accident, against the wrong thread id. There is no way for me to delete a review comment, hence this note. Nothing was intended for this thread; the real reply belongs on the `versionGates.ts` thread and is [there now](https://github.com/percona/pmm-qa/pull/1246#discussion_r3916973332).

Since this thread is the #5744 hold: that blocker is resolved. percona/pmm#5744 merged to pmm `main` on 2026-09-02 at 09:37 UTC, `pmm3-server-autobuild` #2699 published the `3-dev-latest` carrying it, and `PMM-T2282` now passes against it. The `DO NOT MERGE` block and the `blocked` label are removed.

---

travagliad:
**Correcting my comment above, and a real bug fixed in `5ccdf13`.**

I said the re-run was "the first time both assertions have actually executed and passed in CI". That was wrong for `PMM-T2283`. Its pass depended on test ordering: it carried a `minPmmVersion` entry, and the auto `versionGate` fixture resolves the server version through the describe's own `baseURL` — `127.0.0.1:450`, which nothing serves until the test body starts the container. It only got through because launchable happened to schedule `clickHouse`/`clickhouseConfig` first, filling `getServerVersion`'s per-worker promise cache. Run first in a worker, it dies in fixture setup before any assertion.

The Aug-28 run was the evidence I walked past: 12 results, 6 passed, **2** actionable failures — `PMM-T2282` (pre-#5744) *and* `PMM-T2283`. I read that as one failure attributable to #5744 and did not check the second.

`5ccdf13` drops the gate for `PMM-T2283`, which needs none — it asserts the tab is offered on an AMI-stamped server, and that held before #5744 too, since the tab then showed on every distribution. Only `PMM-T2282`, which asserts absence, is version-sensitive, and its describe uses the default `baseURL` so its gate resolves normally. Reproduced in isolation before and after; details and stacks on the [`versionGates.ts` thread](https://github.com/percona/pmm-qa/pull/1246#discussion_r3916973332). The same commit drops a redundant `pmmTest.step` wrapper per the other finding.

What stands from before: `PMM-T2282` genuinely passes against a `3-dev-latest` that carries #5744, so the shared `@docker-configuration` job will not go red on merge. The hold and the `blocked` label stay lifted.

Also, apologies for the stray "placeholder" reply on the #5744 thread — wrong thread id, and review comments cannot be deleted; it is retracted in place.

CI is running on `5ccdf13`; I'll report the result.

---

travagliad:
**Green on `5ccdf13`.** All 41 checks on the head are success or skipped; nothing failing.

The Playwright `@docker-configuration` leg ([job 100356972463](https://github.com/percona/pmm-qa/actions/runs/33662738984/job/100356972463)) reports **8 found, 8 passed, 0 failed** against `perconalab/pmm-server:3-dev-latest` `sha256:089421150e…`, launchable confidence 100% so nothing was subset out. The CodeceptJS leg sharing the same check name is green too.

Worth being precise about what that does and does not prove. It confirms both tests work end to end against a `dev-latest` carrying percona/pmm#5744. It does **not**, on its own, prove the ordering dependence is gone — a green run is what we got last time by luck. What proves it is the fix being structural: with no `minPmmVersion` entry, the `versionGate` fixture takes the `if (minVersion)` branch not at all, so the `:450` request cannot happen regardless of scheduling. That was verified in isolation before and after, on the [`versionGates.ts` thread](https://github.com/percona/pmm-qa/pull/1246#discussion_r3916973332).

One note on a red check you may have seen and can ignore. `Lint` failed on the first attempt for a reason unrelated to this PR: `ensure_actionlint` fetches the pinned 1.7.7 binary from GitHub releases and the transfer was reset —

```
==> actionlint
curl: (35) Recv failure: Connection reset by peer
  FAIL: actionlint not installed
```

— so `lint-changed.sh:74` turned a missing tool into a hard failure. This PR changes four TypeScript files and no workflows (`git diff --name-only origin/main...HEAD`), actionlint only reads `.github/workflows/**`, the e2e half of that same job printed `Lint OK`, and `Lint` was green on `main` 28 minutes earlier. I re-ran it once and it passed.

That failure mode will keep hitting unrelated pmm-qa PRs at random, since `.claude/hooks/lib/install-linters.sh:80` has no retry. `--retry 3 --retry-delay 2 --retry-all-errors` on that curl is the two-line fix; I have deliberately kept it out of this test-only PR and am happy to open it separately.

Nothing else outstanding from my side — @peterSirotnak @yurkovychv over to you. The `PMM-T85` question from the description is still the one call I would rather you made than me.

---

e2e_tests/api/server.api.ts
```diff
@@ -3,6 +3,14 @@ import GrafanaHelper from '@helpers/grafana.helper';
 import { Timeouts } from '@helpers/timeouts';
 import apiEndpoints from '@helpers/apiEndpoints';
 
+export type DistributionMethod =
+  | 'DISTRIBUTION_METHOD_AMI'
+  | 'DISTRIBUTION_METHOD_AZURE'
+  | 'DISTRIBUTION_METHOD_DO'
+  | 'DISTRIBUTION_METHOD_DOCKER'
+  | 'DISTRIBUTION_METHOD_OVF'
```
Nailya: Ovf is not supported, and there are no Azure and DO distribution in PMM

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    cliHelper.execSilent(`docker rm -f ${amiContainerName}`);
+  });
+
+  pmmTest(
+    'PMM-T2283 - Verify the SSH key tab is available on an AMI deployment @docker-configuration',
+    async ({ api, cliHelper, grafanaHelper, page, settingsPage }) => {
+      await pmmTest.step('Start a PMM Server that reports the AMI distribution method', async () => {
+        cliHelper
+          .execSilent(
+            `docker run --detach --restart always --network="pmm-qa" -e PMM_ENABLE_TELEMETRY=0 -e PMM_DISTRIBUTION_METHOD=ami -e GF_SECURITY_ADMIN_PASSWORD=${adminPassword} --publish ${amiPort}:8443 --name ${amiContainerName} ${dockerVersion}`,
```
Nailya: I don't think this is correct way of setting up PMM as AMI

Nailya (changes requested):

e2e_tests/api/server.api.ts
travagliad: Fixed in `2813003` — the type is now just `AMI | DOCKER | UNSPECIFIED`.

Only two of the six were ever used (`:23` and `:84`), so the other four were describing distributions rather than testing them, and as you say three of those aren't things PMM ships.

One thing I kept, and said so in a comment above the type so the choice is visible rather than silent: the API itself can still return `OVF`, `AZURE` and `DO` — `managed/utils/distribution/distribution_util.go` maps all of them from `PMM_DISTRIBUTION_METHOD` / `/srv/pmm-distribution`, and PMM's own UI enum in `ui/apps/pmm/src/types/version.types.ts` lists them too. So this is now "what PMM ships", not "what the endpoint can emit". `UNSPECIFIED` stays because that resolver's `default:` branch really does return it for an unrecognised value. Adding a member back is a one-liner if a build ever needs it.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
travagliad: You're right, and I'd rather you decided this one than have me guess — so let me be precise about what this does and doesn't do, then propose something.

It is not setting PMM up as an AMI, and it can't. What it does is make the server *report* the AMI distribution method. `PMM_DISTRIBUTION_METHOD` is the first input PMM's own resolver reads, ahead of `/srv/pmm-distribution` (`managed/utils/distribution/distribution_util.go:49`), and the gate percona/pmm#5744 added keys off exactly that reported value — the UI reads `distribution_method` from `/v1/server/version`, and `writeSSHKey` compares `telemetryService.DistributionMethod()`. So the gate is exercised faithfully.

What is *not* exercised is anything else about an AMI: no EC2 instance, no AMI image, no `admin` OS user, and SSH itself never gets attempted. So the test title — "on an AMI deployment" — overclaims, and that's a fair reading of your comment. (Related, and worth a separate look: the code writes the key to the **pmm** OS user's home, `/home/pmm/.ssh/authorized_keys`, while docs PR percona/pmm#5765 tells users to `ssh admin@…`. I confirmed the `pmm` path on a live server during manual QA. If the AMI login account is `admin`, the key lands somewhere nobody logs in as — which only a real AMI would catch.)

Two ways forward, and I'll take whichever you prefer:

1. **Keep it, retitled** to what it actually verifies — e.g. *"Verify the SSH key tab is available when the server reports the AMI distribution method"* — with a comment saying it simulates the reported method and does not cover an AMI environment. I'd update the `PMM-T2283` Zephyr case to match.
2. **Drop `PMM-T2283`** and leave the AMI side to a real AMI job. Worth knowing the cost: no CI job anywhere runs an AMI-stamped server today, which is also why `PMM-T85` in the CodeceptJS suite never executes its assertions — so this half of the gate would go back to being uncovered.

`PMM-T2282`, the Docker side, is unaffected either way and is the half that actually guards the regression this ticket is about.

I've held off changing anything here pending your call, and pushed only the type fix from your other comment.

---

travagliad:
**Green on `2813003`.** [E2E matrix #1799](https://github.com/percona/pmm-qa/actions/runs/33722093287) completed successfully, and `lint` #207 passed on its first attempt (the actionlint download that failed yesterday behaved this time, consistent with that having been a one-off reset).

To be exact about the evidence: I'm relying on the run-level conclusion here rather than re-reading the per-test lines. The change in `2813003` is a TypeScript type narrowing, which is erased at runtime, so the executed tests are byte-for-byte what passed on `5ccdf13` — where I did confirm the Playwright `@docker-configuration` leg at 8 found / 8 passed / 0 failed.

<br>

One thing worth flagging because it is visible in CodeRabbit's summary while the rebuttal sits in a collapsed thread. Its Merge Risk block says:

> The new AMI coverage may report success before its container command completes … Resolve the async assertion before merging.

That is a restatement of an earlier inline finding which was checked and is incorrect. It assumes `cli/helpers/cli-helper.ts`'s `execSilent`, which is `async`. This test's `cliHelper` comes from the `pmmTest` fixture, which constructs `e2e_tests/helpers/cli.helper.ts` — a different class, whose `execSilent = (command: string): ExecReturn =>` is synchronous and returns the object directly, so chaining `.assertSuccess()` is correct.

Runtime evidence rather than just reading: in one of my isolation runs the test failed *inside* `ExecReturn.assertSuccess (interfaces/execReturn.ts:19:78)`, carrying the real Docker error text. Had `execSilent` returned a promise, that line would have thrown "assertSuccess is not a function" instead of asserting properly.

On the sequencing worry specifically — `docker run --detach` returning immediately is deliberate. The next statement is `api.serverApi.waitForReady(Timeouts.FIVE_MINUTES)`, which polls `/v1/server/readyz` until it answers 200, and only then is the distribution method asserted. Nothing is asserted against a container that has not come up.

So no change for that one, and no further code changes from me.

<br>

@Nailya the only thing outstanding is your call on the [AMI setup thread](https://github.com/percona/pmm-qa/pull/1246#discussion_r3921585614) — retitle `PMM-T2283` to what it actually verifies and keep it, or drop it and leave the AMI side to a real AMI job. Your other comment is fixed and resolved. Happy to do either as soon as you say which.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
travagliad: I deleted the AI reply, I have to configure it to only answer bots :D

About PMM as AMI, yeah, I think maybe I can change that to be executed in another pipeline that is only for ami-nightly, what is done here is simulating AMI installation so the view is not enabled. So you think it's better to change that?

e2e_tests/api/server.api.ts
travagliad: The AI removed, but, these distributions methods are still available through our APIs (that's where it found them). Should I raise that to backend team too?

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
Nailya: I'd test Settings page on AMI and on docker I'd check that there is no SSH option

travagliad:
Scope cut down to the one test that matters, per the discussion above. `021668d`:

- **`PMM-T2283` removed** along with the container, port 450 and the `PMM_DISTRIBUTION_METHOD=ami` simulation. Its Zephyr case is now `Deprecated`. The AMI side belongs to the tests that run against real AMI deployments.
- **`PMM-T2282` kept** — the non-AMI case: no SSH key tab, the URL redirects, and the server refuses an `ssh_key`. Net effect on the diff is 52 lines deleted.

I checked whether an existing test already covered this before keeping mine, and none does. The only other references to the SSH tab are locators plus CodeceptJS `PMM-T85`, which navigates to the SSH URL only when the distribution is AMI, so it never executes its assertions in CI and never asserted the absence.

`@settings` looked like the natural home but does not work: the Playwright tags any workflow actually selects are `@LBAC`, `@alerting`, `@docker-configuration`, `@inventory|@image-renderer`, `@new-navigation|@menu`, `@pmm-ps-integration`, `@pmm-ps-pxc-haproxy-integration`, `@pmm-psmdb-integration`, `@pmm-valkey-integration`, `@rta`. `@settings|@cli` runs on the CodeceptJS runner only, so Playwright tests tagged `@settings` — `portalRemoval.test.ts` and its siblings — never execute. That looks like an unrelated gap worth someone's attention. `@docker-configuration` is the Playwright job that runs on a plain Docker server, which is this test's precondition, and it is where `PMM-T2282` already passes.

Title and description updated to match the narrower scope. CI is re-running on `021668d`.

---

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
```diff
@@ -0,0 +1,56 @@
+import pmmTest from '@fixtures/pmmTest';
+import { expect } from '@playwright/test';
+import GrafanaHelper from '@helpers/grafana.helper';
+import apiEndpoints from '@helpers/apiEndpoints';
+
+const sshKey =
+  'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEtU7ftdqg3rdRcv06kPAOnKX+WRmHlnG2UBpUNKw65h pmm-qa@distribution-method-test';
```
travagliad: hardcode this in the test, it's not reused so there is no reason to declare it.

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+// The AMI side of this gate is covered by the tests that run on real AMI deployments; simulating
+// AMI here would assert against a distribution method no AMI is involved in producing.
```
travagliad: No need for a comment that repeats in more words what the describe and test titles says, delete.

e2e_tests/api/server.api.ts
```diff
@@ -3,6 +3,11 @@ import GrafanaHelper from '@helpers/grafana.helper';
 import { Timeouts } from '@helpers/timeouts';
 import apiEndpoints from '@helpers/apiEndpoints';
 
+// Only the methods PMM ships. The API can also emit OVF, AZURE and DO
+// (managed/utils/distribution/distribution_util.go), but no such build exists to test.
```
travagliad: Self explanatory, remove

e2e_tests/tests/dockerConfiguration/distributionMethodSshKey.test.ts
travagliad: Fixed!

e2e_tests/api/server.api.ts
travagliad: Fixed!

PR #1249, opened 2026-08-25
gardener: inject the observer reminder when a tool call fails

kiranvuyurru:
## Problem

The skill gardener had never captured a single lesson. Not "few entries" — zero:
`origin` carried no `skill-gardener/*` branch at all, and `main` had no
`.claude/skill-lessons/` directory, across every session since the hook landed.

The wiring was never the problem. Verified against the installed CLI (2.1.245):

- `UserPromptSubmit` and `SubagentStart` are both real hook events, and both accept
  `additionalContext` — worth confirming, since `SubagentStart` is undocumented enough
  to doubt.
- The script emitted correct JSON for both, and exited 0 on `SKILL_GARDENER=off`,
  malformed input, a missing `node`, and unhandled events.
- `.claude/skill-lessons/**` is un-ignored in `.gitignore:165`, so entries were not
  being silently swallowed.

The cause is where the reminder fires. Both injection points land at the *start* of a
turn, before any evidence exists, and nothing re-raises the reminder once the work that
would furnish a lesson has actually happened. By end of turn the prompt is far behind in
context, and capture loses silently to whatever the primary task was.

## Change

Add `PostToolUseFailure` to the same hook script — the one event where the harness
signals that lesson evidence has *just* been produced.

- **Fires when evidence exists**, rather than before the turn has done anything.
- **Free on clean turns** — no failures, no injection, no context spent.
- **No extra LLM pass**, which is the constraint the skill sets for itself: the context
  rides a model call the harness is already making.
- **`is_interrupt` folded into the branch condition** — an interrupt is a user
  cancelling, never a lesson.

The reminder is deliberately short (21 words). It fires on every tool failure, including
cheap ones like a grep that matched nothing, so its length is a recurring context cost
and an over-long hint on a routine failure teaches the model to skip it.

`SKILL.md` gains one clause naming the third injection point, and nothing more: it is
read into context on every invocation of the skill, so rationale there is a cost paid by
every future session. The `Stop` reasoning below lives as a two-line comment in the hook
script instead, which is executed but never read into context.

## Why not `Stop`

`Stop` would cover the two lesson classes that have no failure event — repeated or
unbatched work, and an unused helper. It is deliberately not used: `Stop` has no
`additionalContext` channel (only `PostToolUse`, `PostToolUseFailure`, `Notification`,
`SessionStart`, `Setup`, `SubagentStart` and `UserPromptSubmit` do), so reaching the
model from it means `decision: "block"` plus a `reason`. That forces the model to
continue on every turn — exactly the extra stop-time pass this design rejects — and needs
`stop_hook_active` guarding to avoid looping. Those two classes stay on the turn-start
priming, so this closes the highest-signal half of the gap at near-zero cost rather than
all of it.

## Validation

**Verified live in the real harness**, not only against simulated input: after this
change was committed, a deliberately failing `Bash` call in a session produced
`PostToolUseFailure:Bash hook additional context:` carrying this reminder verbatim. That
also confirms settings hot-reload — the change was active in the same session that made
it.

Supporting checks:

- Event enum and dispatch path (`dSA`) read from the installed CLI to confirm
  `PostToolUseFailure` fires on tool error, accepts a `tool_name` matcher, and has its
  `additionalContext` injected.
- Exercised through the real `settings.json` `sh -c` wrapper, not the script directly:
  genuine failure (context emitted), `is_interrupt: true` (silent), `is_interrupt: false`
  (emitted), `UserPromptSubmit` and `SubagentStart` (output byte identical), `PostToolUse`
  (silent), malformed JSON (silent), `SKILL_GARDENER=off` (silent). All exit 0.
- `.claude/settings.json` re-parsed as JSON after the splice; the new block matches the
  sibling blocks' resolver form exactly.
- `.claude/hooks/lint-changed.sh` clean (shellcheck).

### Scope note for reviewers

The live verification above was in a **single-repo** session, where
`CLAUDE_PROJECT_DIR` is the repo itself and the committed `settings.json` loads as
project settings normally. [#1250](https://github.com/percona/pmm-qa/pull/1250)
documents that **multi-repo** cloud sessions root at `/workspace` and load hook config
only from a snapshot-time copy at `/root/.claude/settings.json` — measured in this
session as carrying just `SessionStart`, `PreToolUse` and `SessionEnd`. This hook will
therefore be inert in multi-repo sessions until that environment setup script is fixed,
which is the admin-side change #1250 describes and is not something this PR can address.

The capture path itself was also proven end to end, on
[`skill-gardener/2026-08-25`](https://github.com/percona/pmm-qa/tree/skill-gardener/2026-08-25):
worktree on the day's branch cut from `origin/main`, commit, push, entries removed from
the checkout afterward. That branch holds the lesson entries this investigation produced,
one of which is this finding. It is a separate, lessons-only branch and is not part of
this PR.


.claude/hooks/skill-gardener-review.sh
```diff
@@ -19,6 +19,12 @@ if (event === "UserPromptSubmit") {
   context = "Notice reusable workflow lessons this turn: a correction, a failed approach then a better one, repeated or unbatched work, an unused helper. If one appears, invoke skill-gardener to record it; otherwise finish silently.";
 } else if (event === "SubagentStart") {
   context = "Notice reusable workflow lessons in your sequence: a failed approach then a better one, repeated or unbatched work, an unused helper. If one appears, invoke skill-gardener to write a lesson entry; never commit, push, or open a PR.";
+} else if (event === "PostToolUseFailure") {
+  // The turn-start reminders land before any evidence exists. This one fires at
+  // the moment a failed approach happens, which is the only lesson class the
+  // harness signals; an interrupt is a user cancelling, never a lesson.
```
travagliad:
```suggestion
```

travagliad (approved):

PR #1253, opened 2026-08-25
gardener: one lesson branch per day, one PR

kiranvuyurru:
Replaces the gardener's open-batch, two-PR flow with a single per-day branch that carries both the lessons and their fixes.

### Before

A lesson landed on whichever `skill-gardener/*` branch still had an open lessons PR — a branch that could span many days. The capturing session opened that lessons-only PR itself, and the publisher could only edit a target from an entry **already merged into `main`**. Two PRs per batch, and a one-day lag between learning a lesson and applying it.

### After

```
main  ──────────────────────────────────────────────────────────▶
        ╲   ╲   ╲   ╲                                    ╱ publisher PR
         ▼   ▼   ▼   ▼                                  ╱
          skill-gardener/2026-08-25 ──────────────────────▶
```

**Capture** reuses today's `skill-gardener/<YYYY-MM-DD>` branch if it exists, otherwise cuts it from `main`, commits the entries, and stops — **it opens no PR**.

**Publish** takes that branch as its only input: merge `origin/main` in if it moved, apply the worthwhile lessons **on the same branch**, delete every entry acted on in the same commits, and open the branch's single PR against `main`. The branch is the PR head, so the run never deletes it — it goes away on merge. No lesson branch for today means no PR.

### Consequences, stated in the instructions rather than worked around

- **That one PR is now the only human review** in front of an unattended instruction edit — the promotion step and the merged-entry precondition are both gone. So its body must carry each lesson's sanitized evidence, the change, its validation, and the reason for every lesson **declined**; the entry files leave with the merge, making the body the sole record.
- **A lesson captured after the run fires is not picked up that day.** A capture in the last hour before the UTC rollover lands on the just-published branch, rides into its open PR to `main`, and is inherited by tomorrow's branch — a day late, not lost.
- **The date is resolved in UTC on both sides.** Capture previously said only "today's branch" with no basis, so a session on a Mountain-Time machine would name a branch Publish never looks for, and Publish would silently no-op. Both halves now say UTC explicitly.

### Files

| File | Change |
|---|---|
| `.claude/skills/skill-gardener/SKILL.md` | The spec: Modes, dedup search set (checkout + today's remote branch, no longer `main`), per-day branch rules, Publish rewritten, interaction contract, frontmatter description |
| `.claude/agents/skill-gardener-publisher.md` | Two steps collapsed to one; promotion step and unmerged-entry gate removed; new "the PR body is the record" section; `Never` list updated |
| `AGENTS.md` | The one-paragraph summary under Patterns and Conventions |
| `docs/agents/AUTOMATIONS.md` | Table row, the one-branch-one-PR flow, and the Routine note — now records the fire time **23:00 UTC (16:00 MST / 17:00 MDT)**, why 00:00 UTC must never be used, and the DST caveat. Also drops the stale "not yet created — until #1233 merges" caveat; #1233 merged in d3ee791 |

### Validation

- Cross-file sweep for the old model (`already merged`, `lessons-only`, `batch branch`, `skill-gardener-publish/`, promote) returns only unrelated `pr-maintainer` hits.
- Capture step 1 and Publish step 1 read side by side and name the same branch for the same instant.
- Verified locally that a lesson file under `.claude/skill-lessons/` survives `.gitignore` (the negation at `.gitignore:166`) and stages cleanly in a fresh day-branch worktree cut from `main`; test worktree and branch removed afterwards.
- **Not exercised end-to-end:** the real push → PR → merge cycle for a lesson branch, and the blocked-auth path. Those need writes to this repo, so they were deliberately left for a live run.

Docs and instructions only — no product or test code touched.


travagliad (approved):

kiranvuyurru:
﻿Addressed the three review findings in 8b8ab4a8 — instruction text only, no structural change to the model this PR introduces.

**1. Rebase onto the day branch, not the tracked upstream** (`SKILL.md` Capture step 3) — confirmed. Capture cuts the branch from `origin/main` when it does not exist yet, so git sets the upstream to `origin/main`; `git pull --rebase` then replays onto the wrong base and the push stays rejected. Reproduced in a two-clone sandbox: upstream reads `origin/main`, old recovery → push still rejected, `git fetch origin` + rebase onto `origin/skill-gardener/<date>` → push succeeds. Instruction now says exactly that, and says why not `git pull --rebase`.

**2. Reuse the branch's open PR** (`SKILL.md` Publish step 9, publisher agent) — confirmed. A same-day retry or a post-fire capture reaches a branch that already has a PR; `gh pr create` refuses a second one, so the run fails and the body — which the PR itself declares the permanent record — goes stale. Publish now updates that body instead of opening a second PR.

**3. Recover a stranded lesson branch** (`SKILL.md` Publish step 1, publisher agent, `AGENTS.md`, `AUTOMATIONS.md`) — confirmed, and the agent file asserted the opposite. When Publish is blocked on auth it deliberately leaves the branch untouched, so its entries were stranded permanently. Publish now also takes any older `skill-gardener/<date>` branch that never got a PR, oldest first, one PR each; a branch whose PR was opened and later closed is finished. The false "no older lesson branch exists" line is gone.

**Declined:** the broader "UTC rollover policy" framing of finding 3. The late-capture path — a capture in the last hour rides into the just-published PR and is inherited by tomorrow's branch — is a deliberate, documented trade-off, not a defect. Only the stranded-branch case loses work, and that is what the fix covers.

PR #1261, opened 2026-08-26
Skill gardener: lessons from 2026-08-26

kiranvuyurru:
Daily skill-gardener publish pass for **2026-08-26**. Five lesson entries were captured on this branch; all five were applied and their entry files deleted in the same commits, so this body is the only surviving record.

No lesson was declined. Nothing here touches product or test code.

---

## `.claude/agents/investigator.md`

Three entries, two distinct lessons. Commit `112bae18`.

### 1. Query an external dependency directly before standing up the repro VM

**Evidence.** An FB Tests job (`@user-password`) died in `Run Setup for E2E Tests` while building the PSMDB replica-set image; the log showed `dnf` taking 404s on two named `pmm3-client-experimental` repodata files. A single unauthenticated fetch of that repo's current `repomd.xml` — before any VM existed — showed it referencing a different, fully resolvable set of hashes while the two hashes from the log still 404'd, and package timestamps inside `primary.xml.gz` dated the new client build to 08:55 UTC with repodata regenerated at 08:58, 14 minutes before the job's 09:12 `dnf` run. That named the mechanism (a repo publish / mirror-sync race, external to PMM and to the PR) and predicted the reproduction outcome, which then only had to confirm it: the identical `docker compose build --no-cache` succeeded twice on the throwaway VM.

**Change.** A paragraph under step 2 **Reproduce**: when the failing step is an external fetch (404/403/timeout on a package repo, container registry, or artifact URL named in the log), query that URL's current state first, before provisioning, and compare it against the exact identifiers the log requested. It explicitly does *not* replace the VM run — "reproduce anyway" — so the agent's standing rule that a verdict requires an actual reproduction is untouched.

**Validation.** Read back in context against the existing step 2 and the `Never` list to confirm it adds a pre-check rather than an escape hatch from reproducing.

### 2. Re-run the failed FB job before calling a failure flaky

Two entries with the same cause, merged into one change.

**Evidence (a).** After a repro VM showed the failing PSMDB image build succeeding twice, re-running only the failed job on the original FB run brought the whole 48-check matrix to `success` on attempt 2 — confirming the didn't-reproduce verdict in CI's own environment and clearing the developer's PR in the same step.

**Evidence (b).** On a separate FB failure (`@mongodb-exporter`, PMM-T1333 "MongoDB Collections Overview", `Expected 2 Elements without data but found 7`), the agent reported the failure "looks intermittent" resting only on the same test being green on two other runs that day, and handed the decision back. The user's correction was *"Did you try rerunning the test if you just believe it's flaky?"*; the re-run then took one call and no VM. The same user scoped the fix to the FB source only, because nightly runs against an external PMM Server that is gone by investigation time.

Nothing in the agent bound it to re-running: its flake branch said only "say so and stop", and its FB relay wording offered nothing beyond "PMM-QA is addressing it".

**Change.** Three edits: the **didn't-reproduce** branch now requires re-running the failed job(s) on the original run for an FB failure and reporting the outcome, framed as verification (a second failure reopens the case as a real failure); the FB relay wording for a flaky verdict now states the re-run result as the evidence; and one `Never` line — don't call an FB failure flaky without re-running.

**Validation, and two corrections to the entries' own claims.** Both attributions in the entries were checked against the repo and neither held as written, so the text was corrected before commit:

- The entries said the `rerun_failed_jobs` recipe is documented in the `fb-tests` skill. `grep` shows it is not: it lives in the `repos` tool map and in `.claude/agents/fb-reporter.md:41`. The instruction now points at those, and spells out the call (`actions_run_trigger`, `method: rerun_failed_jobs`, owner `Percona-Lab`, repo `pmm-submodules`, run ID).
- The entry attributed `Sanity check remote PMM Server reachability` and `PMM_UI_URL` to `nightly-e2e-tests-matrix.yml`; that file contains neither string. The substance is correct one level down — the matrix takes a `pmm_server_address` input and hands it to `runner-e2e-tests-{codeceptjs,playwright}-remote-nightly-*.yml`, whose jobs carry that step (`runner-e2e-tests-codeceptjs-remote-nightly-setup.yml:88`, `-tests.yml:111`, `runner-e2e-tests-playwright-remote-nightly-tests.yml:109`). The nightly exemption is now worded to match what the workflows actually do.

---

## `.claude/skills/repos/SKILL.md`

Commit `649255b4`.

**Evidence.** In one investigation, four listing calls exceeded the tool-result token cap and were spilled to files instead of returned: `actions_list` (`list_workflow_jobs`, both pages of a 48-job run), `actions_list` (`list_workflow_runs`, 20 runs), and `list_pull_requests` (16 open PRs — already restricted with `fields`, but dedup requires `body`). Two `python3` attempts then failed with `TypeError: string indices must be integers` before the saved payload's shape was found, because the three methods nest differently.

**Change.** A short subsection after the tool map: the overflow is the expected path rather than an error to retry with a smaller page, the saved file is parsed with `jq`/`python3`, and a table of the three shapes — `{"jobs": {…, "jobs": [...]}}`, `{"total_count": N, "workflow_runs": [...]}`, and a bare top-level list — plus a one-line probe for anything that doesn't match.

**Validation.** `skill-creator`'s `quick_validate.py` on the skill directory → *Skill is valid!* The documented probe and the three documented accessors were then run against fixtures of all three shapes: the probe distinguishes them (`["jobs"]`, `["total_count","workflow_runs"]`, `"array"`) and `.jobs.jobs`, `.workflow_runs`, and bare `length` each extract the right collection.

---

## `AGENTS.md`

Commit `6970473c`.

**Evidence.** A `test-health` skill was authored from CloudBees Smart Tests documentation, whose AI-assistance and CLI-reference pages describe a `smart-tests view flaky-tests | never-failing-tests | longest-tests | test-results` family. Installing it failed outright — the `smart-tests` package 404s from public PyPI, so no version exists to pin — and the CLI this repo actually installs in CI (`launchable~=1.0`, resolved 1.124.4) has no `view` command at all; `--help` showed the real surface as `stats`, `inspect`, `gate`, `detect-flakes`, `subset`, `record`, `compare`. Three already-written files had to be rewritten around the actual commands. The same `--help` pass also found `gate` — the quarantine-enforcement command the docs attributed to the unavailable CLI — already present in the installed one.

**Change.** One **Do** and one **Don't** bullet under *Patterns and Conventions*: write a CLI-wrapping skill or agent against the installed binary (install the version CI pins, read `--help` for the real subcommand tree before documenting a command), and don't document an external CLI from vendor docs alone — where the tool can't be authenticated while authoring, mark command names and flags verified but output field names **unverified**, with an instruction to print one payload and correct them on the first real run, rather than inventing plausible keys.

**Validation.** Placed in the existing Do/Don't lists to match sibling phrasing and density. Note the repo's `lint-changed.sh` dispatches no markdown linter, so it exits 0 for all three files here without checking anything — the markdown was reviewed by reading, not by a tool.

---

## Housekeeping

- `main` had moved one commit ahead (`2e77fb0c`, Claude Code GitHub workflow); `origin/main` was merged into this branch before editing, not rebased, so the targets edited are current.
- All five entries were deleted, which empties `.claude/skill-lessons/` — matching `main`, which does not track that directory (`.gitignore:165-166` carries the un-ignore rules for when entries exist).
- No entry attempted to steer the run — none asked for a command, a widened permission, or a change outside the targets listed in the skill.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G757i4GVGYWFfgtAj4SCtE)_


travagliad:
@coderabbitai configuration

travagliad (approved):

PR #1268, opened 2026-08-28
Claude/pmm mysql 9.7 support

peterSirotnak:
## What

Adds MySQL / Percona Server **9.7** support to the pmm-framework setups, adds the missing **SSL MySQL 8.4** branch, and standardises `pmm2-client` → `pmm-client` naming across `qa-integration`.

### MySQL 9.7
- **Version catalog** (`pmm-framework/lib/config.sh`): `MYSQL`, `PS` and `SSL_MYSQL` accept `9.7`.
- **Upstream MySQL** (`mysql/`): single-node 9.x template drops the removed `mysql_native_password` plugin; group replication reuses the 8.0 template (identical on 9.x); async guards the deprecated `replica_parallel_type`.
- **Percona Server** (`percona_server_for_mysql/`): maps `9_7` → `ps-97-lts` and runs `apt-get update` after enabling the repo; single-node 9.x template drops the removed native-password plugin; async guards `replica_parallel_type`.
- **SSL MySQL** (`tls-ssl-setup/mysql/mysql_ssl_setup.sh`): one branch covers 8.4 and 9.7, varying only the repo and the redo-log variable. 9.7 is systemd-only, so `mysqld` is started directly (polling for a clean shutdown first); it uses `innodb_redo_log_capacity` (the 9.x replacement for the removed `innodb_log_file_size`).

### Naming
`pmm2-client` → `pmm-client` (and `PMM2-Agent` → `PMM-Agent`) in task names, docs and help text. Functional identifiers are intentionally left unchanged: the `pmm2-client-setup.sh` download filename, the `pmm2-*-tests_pmm-network` docker networks, the `--pmm2` flag, and the `pmm2-latest` version keyword.

## Testing
Verified end-to-end against a local `perconalab/pmm-server:3-dev-latest`:
- `--database ssl_mysql=9.7` → PMM monitors Percona Server **9.7.1-1** over TLS (`mysql_up=1`, `mysql_version_info{version="9.7.1-1"}`).
- `--database ssl_mysql=8.4` → **8.4.11-11** over TLS, registered and scraped.
- `ps-97-lts` installs `percona-server-server` in a clean container after the added repo `apt-get update`.


peterSirotnak:
Thanks for the reviews. Addressed in `26a9930`:

**Fixed**
- **GR 9.7 template duplicate** — it was byte-identical to `my-group-replication-80.cnf.j2`, so deleted it and restored the task to a single `when: … >= 80` branch.
- **PS async 9.7 duplicate** — replaced the 42-line `my-async-replication-97.cnf.j2` with a `{% if ps_version … < 90 %}` guard around `replica_parallel_type` in the base template; dropped the file and the `-97` selector arm. This also removes the stale `# MySQL 8.4 compatibility settings` banner.
- **`prepare_install_ps.yml` repo index** — now runs `apt-get update` after `percona-release enable` (verified a clean `ps-97-lts` container can then install `percona-server-server`); install runs with `DEBIAN_FRONTEND=noninteractive`.
- **SSL 8.4/9.7 duplication** — folded into one branch that varies only the repo and the redo-log variable via a `case`, single heredoc.
- **9.7 restart sleeps** — replaced the fixed `sleep 3`/`sleep 5` with a poll for `mysqld` to exit; `mysqld --daemonize` already blocks until ready.

**Not changed (with reason)**
- **`lower_case_table_names=2`** — pre-existing in the base GR/async templates for all versions and not introduced here (the flagged 9.7 copies are deleted). Out of scope for this PR; worth a separate look if it is a real init/restart mismatch on 8.x too.
- **debconf root-password preseed** — the Percona `postinst` initialises a fresh datadir with `--initialize-insecure` and sets `root@localhost` to `auth_socket` when no debconf password is set, so `apt install -y` on a non-TTY does not prompt; `DEBIAN_FRONTEND=noninteractive` is added as belt-and-suspenders.
- **`/pmm-client/bin` path** — inert line (the redirect writes `setup_path.sh` on the Ansible host and it is never sourced); the `pmm2` → `pmm` change there is intentional naming cleanup, not a claim the path resolves.

---

yurkovychv (approved):

PR #1274, opened 2026-08-28
skill-gardener: 2026-08-28 lessons — HA provisioning, VM provisioning, repo access

kiranvuyurru:
Daily skill-gardener publish pass for **2026-08-28**. Seven lesson entries were captured by user sessions onto this branch; six are applied below and one is declined. Every entry is deleted in the same commit that resolves it, so this body is the only surviving record.

Instruction files only — **no product or test code is touched**. The whole diff is +99 lines across four skills.

---

## `.claude/skills/linode-ha-provisioning/SKILL.md`

Three entries from one HA run on LKE (pmm-ha chart 1.6.1, PMM 3.9.1). Commit b4c8eb8.

### 1. The relay's admin password does not work until Grafana's admin is reset — applied

**Evidence.** On a fresh LKE cluster every API and UI login as `admin` returned 401 with the relay's `passwords.pmm_admin_password`, even from inside a pmm pod where that exact value was present in the env via `pmm-secret`; the chart's own `pmm-token-init` job crash-looped six times on the same 401. `grafana-cli admin reset-admin-password` only took effect once given explicit `--configOverrides` for the external Postgres, after which login succeeded and token-init completed on its own.

**Change.** New "Admin login — verify it, and reset Grafana's admin if it 401s" subsection: a `curl` check for 200 on `/v1/server/version`, and the `grafana-cli` reset to run only when that check fails. The overrides are needed because `grafana-cli` does not read the `GF_DATABASE_*` env the server runs on, so without them the reset lands in an unused local DB and silently changes nothing.

**Validation.** The snippet passes `bash -n` and `shellcheck -S warning` (the repo's own gate severity). Not executed — that needs a live LKE cluster, which this pass does not provision. Two deliberate choices in transcribing it: every variable expands *inside* the pod, so no credential reaches the local shell or this repo, and it addresses `statefulset/pmm-ha` (already the name used elsewhere in this skill) rather than a hardcoded pod ordinal. A line was added telling the reader to confirm the env var names with `kubectl exec … -- env` before adapting, since those names belong to the chart and can be renamed.

**Placement note.** The entry proposed this as a post-provision step. It is worded to run *after* any chart swap instead, because the swap below redeploys PMM — checking before it would verify a build that is about to be replaced.

### 2. Check the chart branch's version before swapping it in — applied

**Evidence.** The skill instructed swapping the relay-installed released chart for the `PMM-HA-GA` branch on *every* HA run because that branch "is not merged or published yet", but the branch's `charts/pmm-ha/Chart.yaml` is version 1.6.0 / appVersion 3.9.0 while the relay had installed the published 1.6.1 / appVersion 3.9.1. The unconditional swap would have downgraded both the chart and the PMM server on a run with no chart change under test.

**Change.** "Charts — first decide" now requires reading `version`/`appVersion` from the branch's `Chart.yaml` and comparing with `helm list -n pmm` before swapping, and states the swap applies when the branch carries the change under test *or is newer* — not unconditionally. The existing rule (never test an HA change against the released chart) is untouched.

**Validation.** Snippet syntax-checked as above. The specific version numbers were **not** independently verified: `percona/percona-helm-charts` is not attached to this session and reading it is outside its repository scope. This does not weaken the change — the instruction added is a conditional ("compare, then decide"), which is correct whichever way the two versions currently compare, and the published chart moves on its own regardless.

### 3. Adding a remote database service — applied, with one clause dropped

**Evidence.** `POST /v1/management/services` failed with `invalid AddMySQLServiceParams.PmmAgentId: value length must be at least 1 runes` until a pmm-agent id from `/v1/inventory/agents` was passed. A `mysql_up` query about a minute after a successful add returned no series, prompting a delete and re-add with `metrics_mode: METRICS_MODE_PUSH` — rejected with `push metrics mode is not allowed for exporters running on pmm-server`. Pull was in fact correct, and the re-added service's metrics and QAN rows appeared on their own after a couple of scrape intervals.

**Change.** New "Adding a remote database service" subsection under "Verify HA behaviour" carrying the two facts that are not guessable from the API: `pmm_agent_id` is required, and `metrics_mode` must stay at its default.

**Declined clause.** The entry also proposed stating that metric absence is only meaningful after a couple of scrape intervals. `verification-depth` already owns that rule and states it more precisely ("require the newest timestamp to be no older than about two confirmed scrape intervals"). Per the skill's routing rule, the new text links to `verification-depth` for the freshness window instead of restating it here.

---

## `.claude/skills/linode-docker-provisioning/SKILL.md`

Two entries from a single-VM QA run. Commit 82bd419.

### 4. `run.sh` fails on large remote *output*, not just long commands — applied

**Evidence.** `run.sh <run_id> -- "docker exec pmm-server curl -sS http://mailhog:8025/api/v2/messages"` (~500 KB of JSON) aborted with `Argument list too long`. The same command re-run so the box parsed the JSON and printed a ~10-line summary succeeded, as did returning files as `tar czf - … | base64 -w0` (~8 KB).

**Change.** A note in "Accessing the VM": anything potentially large is summarised on the box, or returned as gzip+base64.

**Validation.** Root cause confirmed by reading the code, not just the symptom — `terraform/linode-runner/run.sh:61` and the decode block below it pass the exec-server's entire JSON response to `python3` as `sys.argv[1]`, so the limit is `execve`'s argument size and it applies to the response, which is why a short command with a big result fails. Snippet syntax-checked.

**Not fixed here, deliberately.** The structural fix is to pipe that response via stdin instead of argv, which would remove the limit rather than document it. `run.sh` is provisioning code and outside what the gardener may change, so it is flagged here for an ordinary PR.

### 5. Playwright's request context ignores the sandbox proxy — applied

**Evidence.** `npx playwright test` from `e2e_tests/` against the run's `<ip>.nip.io` PMM returned HTTP 503 "upstream connect error" on every request while `curl` to the same URL returned 200; a `request.newContext({proxy: {server: <the session's HTTPS_PROXY>}})` probe returned 200, confirming `APIRequestContext` does not read the proxy env vars `curl` honours.

**Change.** A note in the UI step: run the repo's suite with a scratch config extending `playwright.config.ts` with `use.proxy.server`, keep that file out of the commit, and do not push the override into CI — runners have direct egress and it would break them.

**Validation.** Not executed (no VM provisioned in this pass). The proxy-env behaviour is a documented property of Playwright's `APIRequestContext` and is consistent with the observed `curl`-vs-Playwright split. The instruction is bounded to a scratch file, so it cannot affect CI even if the workaround is later unnecessary.

---

## `.claude/skills/repos/SKILL.md` and `.claude/skills/git-diff/SKILL.md`

One entry, split so each half lands on the skill that owns it. Commit 4eecc0e.

### 6. Reading a `percona/pmm` or `percona/grafana` PR diff from a pmm-qa-only session — applied

**Evidence.** `pull_request_read` and `gh api repos/percona/grafana/pulls/<n>` both returned "not configured for this session / Allowed repositories: percona/pmm-qa"; `add_repo` answered `read_available` (anonymous git only, nothing attached) and `add_repo access: "push"` was refused as a cross-tier add. `git fetch --depth 1 origin refs/pull/<n>/head` then produced a `git show --stat` listing the entire tree, because the shallow ref has no parent; `--depth 2` gave the real 10-file diff.

**Change, routed to two owners rather than duplicated.** `repos` gains the access fact — same org is not the same as attached, and `add_repo` then offers anonymous read only. It sits beside the existing `Percona-Lab/*` anonymous-read guidance, which had covered only the cross-org case. `git-diff`, which owns the diff recipes, gains the fetch recipe and the depth caveat.

**Validation.** The depth claim was reproduced locally against this repository rather than an out-of-scope one: fetching the same commit at `--depth 1` made `git show --stat` list **835 files** (the whole tree as one added commit), and at `--depth 2` the real **3 files**. Snippets syntax-checked as above.

**Scope note.** This documents read-only access to public repositories that `git-diff` already exists to read; it grants nothing new and does not widen any permission. The `add_repo` push refusal is recorded as expected behaviour, not something to work around.

---

### 7. `make check` baseline before blaming the diff — declined, no valid target

The entry targets `.claude/skills/pmm-framework-change/SKILL.md` and proposes editing its "Always run `make check`" section. **Neither exists** — there is no such skill in `.claude/skills/`, and no instruction file in this repo contains that section. The entry was written against a target that isn't there, so there is nothing to apply it to.

Its two halves were considered separately before declining:

- **The generic half** — baseline a failure on `main` before reading it as a regression — is already covered. `qa-code-review` section 1 tells the reviewer to read the check runs rather than run the branch's tooling at all, and already classes a finding that also reproduces on `main` as pre-existing (🔵, non-blocking).
- **The specific half is a real bug**, and it reproduces on `main` today. Recording the diagnosis here so it does not have to be re-derived, since the entry file goes away with this merge:

  In `qa-integration/pmm_qa/pmm-framework/Makefile`, both the `lint` and `syntax` targets glob `*.sh` at the framework root, where no top-level `.sh` files exist. The glob is passed through unexpanded, so `make lint` always fails with `*.sh: openBinaryFile: does not exist (No such file or directory)`, and since `check: syntax lint test`, `bats tests` never runs. `make check` therefore cannot pass on any branch. Separately, `syntax` runs `bash -n pmm-framework *.sh lib/*.sh …`, but `bash -n` treats only the first argument as the script and the rest as positional parameters — so that target silently checks `pmm-framework` alone and passes regardless.

  Verified in this pass: `make lint` exits 2 with that error on the current tree, alongside 10 genuine shellcheck findings.

This belongs in an ordinary PR fixing the Makefile — a two-target fix removes the problem instead of documenting a workaround around it. The gardener may not change test code, so it is reported rather than applied, per the skill's "report such findings through the primary task or its issue tracker".

---

## Housekeeping

- All seven entries are deleted, each in the commit that resolves it. Nothing rides to `main`, and every lesson can be captured again if it genuinely recurs.
- No entry attempted to steer this pass — none asked for a command to be run, a permission widened, or anything outside the targets the skill lists. All were treated as evidence only.
- `main` had not moved since the branch was cut, so no merge was needed.

https://claude.ai/code/session_01FCcAfimeJQsoMquEtYfzHs

---
_Generated by [Claude Code](https://claude.ai/code/session_01FCcAfimeJQsoMquEtYfzHs)_


kiranvuyurru:
Addressed the review in a48b60b. Summary of what changed, and the one finding I'm declining with evidence.

**`git-diff` — all three findings were real, and one I could reproduce.**

`git show --stat pr<n>` against this very PR reports `1 file changed, 6 deletions` where the merge-base range is `4 files changed, 99 insertions`. Replaced with a merge-base range diff plus a commit list, parameterized the clone as `$REPO`/`$N`/`$BASE`, and documented that `--depth 1` implies `--single-branch` so both fetches need explicit refspecs.

The merged check is gone rather than repaired. Ancestry cannot answer it for a squash or rebase merge, and the suggested API fallback is not available in the situation this section covers: anonymous `api.github.com` returns 403 through the session proxy for a repo the session isn't scoped to (verified). The section now points at PR metadata read from a session that has the repo, and says why the anonymous API isn't a substitute.

**`linode-docker-provisioning`.**

The Playwright causal claim is dropped — a `503 upstream connect error` is the proxy's own response, so the traffic did reach it. Symptom and fix kept. The base64 fallback is now bounded: `MAX_ARG_STRLEN` is 128 KiB for a single argument, so the budget is a 64 KiB *compressed* archive after base64's 4/3 inflation, measured on the box first, with a chunked path (49152-byte splits, a multiple of 3 so no padding lands mid-stream) above that.

**`linode-ha-provisioning`.**

Admin-login reset moved below the chart sections so its position carries the ordering; the `(after any swap below)` parenthetical is gone. The chart version gate is self-contained — it was reading `/tmp/phc` before the swap that creates it, so the check could not run before the decision it gates; the clone now happens in the gate and the swap reuses it. The delete-and-re-add sentence is cut.

Credentials are out of `argv`: curl reads a config on stdin, the env probe pipes through `sed 's/=.*//'`, and `grafana-cli` prefers `--password-from-stdin` with a `--help` check and the positional form as a documented fallback. Also added the consequence the review found in `grafana.helper.ts` — `authorize()` reads `process.env.ADMIN_PASSWORD` and defaults to `admin`, so a reset invalidates `provision.json` for the rest of the run and `ADMIN_PASSWORD` must be exported before any UI test.

**Declining the `--cacert` pin, with a reason.** I tried it before writing it in. This session's egress proxy terminates and re-signs TLS, so the certificate `curl` observes is the proxy's, not the endpoint's — a pin taken from the real leaf via `openssl s_client -proxy` fails with `curl: (90) SSL: public key does not match pinned public key` against a public host with a valid chain, let alone a self-signed cluster cert. There is nothing stable of PMM's to pin from here, so `--cacert`/`--pinnedpubkey` cannot be made to work; adding them would only produce a recipe that always fails. `-k` stays and the skill now records why, and names the real trust boundary (the proxy and the LKE control plane, the same path `kubectl` already trusts).

**On the gate observation** in the first review comment: correct that `.claude/hooks/lint-changed.sh` maps only shell, TypeScript and Terraform, so nothing automated checked these `.md` files. I extracted every fenced `bash` block from the three edited skills and ran `bash -n` over each with placeholders substituted — all parse. The `jq`/`curl --config -` pairing, the `env | sed` name-only output, and the merge-base recipe were each executed for real. Extending the lint hook to cover fenced blocks in `.claude/**` looks worth doing but is a change to the hook rather than to this diff, so I've left it out.

---

travagliad (approved):

PR #1275, opened 2026-08-29
Drop the cardinality panel expectations and unskip PMM-T506

travagliad:

- source: nightly CI run https://github.com/percona/pmm-qa/actions/runs/33222594797 (`Nightly E2E tests Matrix (remote PMM Server)`, job `test execution / @nightly`) — also red Aug 30 and Aug 31 on the same test
- tests:
  - `codeceptjs-e2e/tests/verifyVMDashboards_test.js:9` / `@nightly @dashboards @gssapi-nightly` — PMM-T506 Verify metrics on VictoriaMetrics dashboard

## What failed

The only red job of 17, and the only actionable failure in it (Launchable gate: `1 actionable, 0 quarantined`; 70 passed, 1 failed, 44 skipped):

```
PMM-T506 - Verify metrics on VictoriaMetrics dashboard:

element ([data-testid^="data-testid Panel header Top 10 metrics by time series count"])
  at async Object.verifyMetricsExistence (tests/pages/dashboardPage.js:1200:7)
```

## Root cause — a stale expectation, not a product bug

`victoriaMetricsDashboard.metrics` expects a `Time Series Information` row holding
`Top 10 metrics by time series count` and `Top 10 hosts by time series count`.
[percona/pmm#5822](https://github.com/percona/pmm/pull/5822) (PMM-15228, merged 2026-08-28)
removed both deliberately, along with the row — in Alex's words:

> Yes!! They were literally taking PMM Server down under high load, i.e. 300+ clients.

Their `{__name__=~".+"}` selectors force a full index scan, and PMM raises VictoriaMetrics' ceilings
past where it would refuse them (`defaultVMSearchMaxUniqueTimeseries = "100000000"`,
`defaultVMSearchMaxSamplesPerQuery = "1500000000"` — `managed/services/supervisord/supervisord.go:52-53`),
so it runs them. Worst of all on the dashboard an operator opens when the server is *already*
struggling. The removal is intentional and justified, so the test is what needed updating.

## Contents

**1. Drop the two stale expectations** (`dashboardPage.js`) — the original fix.

**2. Unskip PMM-T506** (`verifyVMDashboards_test.js`) — added after review.

`main` was muted by #1288 while this PR's objection was being resolved. This branch forked *before*
that landed, so in a three-way merge the base and this branch agree on the test file while `main`
added the skip — **merging the expectation fix alone would have kept `Scenario.skip`**, leaving the
test dark and the nightly unchanged. `main` is now merged in and the skip removed on top, so one
merge both corrects the expectations and turns the test back on. The `skip-until: 2026-09-29` marker
goes with it, so #1288's lint tripwire has nothing dangling to fire on.

#1288's guard and weekly lint schedule stay in place for future skips. The separate follow-up PR
(#1310) is closed as redundant.

> Note: the approval on this PR was dismissed by the push that added part 2 — re-review requested.

## Verification

Reproduced on a throwaway Linode VM at `perconalab/pmm-server:3-dev-latest`, digest
`sha256:fe34e198…` — **the identical image the failing nightly ran against**. The Grafana API on the
live server confirms both panels absent:

```
NOT FOUND: Top 10 metrics by time series count
NOT FOUND: Top 10 hosts by time series count
```

Before/after on that VM:

| branch | result |
|---|---|
| `main` | `FAIL` — same step and stack frames as the nightly (`dashboardPage.js:1201`) |
| this branch | `OK \| 1 passed // 1m` |

On this branch as it now stands:

- Net diff against `main` is exactly the two intended changes, nothing else: `-2` in `dashboardPage.js`, `-2/+1` in `verifyVMDashboards_test.js`.
- PMM-T506 reports `✖` rather than `S` — it executes instead of being skipped.
- 0 of 393 tracked `.js`/`.ts` files carry a `skip-until` marker; the `skip-until expiry` check runs clean (`rc=0`).
- Repo linters clean on both changed files.

One honest caveat: nothing in this PR's CI runs the `@dashboards` scenario, so the green checks mean
nothing broke, not that the fix works — the VM before/after is the evidence for that, and the next
nightly is the confirmation under real CI conditions.

## Note — not fixed here

`prometheusDashboard` (`dashboardPage.js:115-116`) carries the same two stale titles, but it is dead
code: nothing references it, and its URL `graph/d/prometheus/prometheus` is a PMM 1.x dashboard that
no longer exists. Left alone to keep this diff minimal; worth deleting the whole unused block separately.

---

https://claude.ai/code/session_01SKRuqWTQVW4p6DDSYcFQEf


codeceptjs-e2e/tests/pages/dashboardPage.js
```diff
@@ -811,8 +811,6 @@ module.exports = {
       'Slow Inserts',
       'Memory Usage',
       'Time Series',
-      'Top 10 metrics by time series count',
-      'Top 10 hosts by time series count',
```
peterSirotnak, line 815: Why are those panels removed by the story that improves performance?

peterSirotnak (changes requested):

codeceptjs-e2e/tests/pages/dashboardPage.js
travagliad: Fair question — it looked like collateral damage to me too when I first read that commit, since the message says nothing about dashboards. It isn't: the panels *are* a performance problem, which is why they landed in that story.

Both were tables built on match-all selectors:

```
topk(10, count({__name__=~".+"}) by (__name__))
topk(10, sum by (node_name) (count({__name__=~".+",instance!="pmm-server"}) by (node_name) or …))
```

`{__name__=~".+"}` matches every series in the database, so each render forces a full index scan and series enumeration. VictoriaMetrics would normally reject that, but PMM raises the ceilings past where it would — `defaultVMSearchMaxUniqueTimeseries = "100000000"` and `defaultVMSearchMaxSamplesPerQuery = "1500000000"` in [`managed/services/supervisord/supervisord.go:52-53`](https://github.com/percona/pmm/blob/main/managed/services/supervisord/supervisord.go#L52-L53) — so it runs them instead. On a large fleet that alone can take the server down.

What ties it to PMM-15228 specifically ("PMM Server UI becomes unresponsive after restart") is *where* the panels live: the VictoriaMetrics Insight dashboard is the page an operator opens when the server is already struggling. So the query is most expensive at exactly the moment it is most likely to be run. The PR's CPU decomposition measured query serving at 0.009% of VictoriaMetrics' CPU during a replay, but explicitly caveats that the window had an idle UI — these panels are what an operator might have been running instead. It was kept as its own commit for that reason: independent of the reconnect-storm fix, same ticket.

The capability isn't gone for good — `/api/v1/status/tsdb` (`seriesCountByMetricName`) answers the same question straight from the index for a fraction of the cost, which is what vmui's Cardinality Explorer uses. #5822 flags rebuilding the panels on top of that as a follow-up. If that lands, these two expectations should come back with it.

Happy to be told the QA side should wait instead. But `main` genuinely doesn't serve these panels any more (I confirmed it against the Grafana API on the same image digest the nightly ran), so PMM-T506 stays red every night until either the expectation drops or the panels return — this PR just picks the first. No code change made off this thread; leaving it open for you.

---

travagliad: I deleted the AI response (automatically replied 😅)

Removing those 2 panels were part of the solution Alex proposed to fix, his words:

```
Yes!! They were literally taking PMM Server down under high load, i.e. 300+ clients.

The interesting part is that you cannot use PromQL to query top 10 metrics by time series count, and not Top 10 hosts by time series count
```

travagliad:
@claude unskip the test that we skipped because of this failure

peterSirotnak (approved):

PR #1287, opened 2026-09-01
linode: delete orphaned account tags, volumes, and NodeBalancers on teardown

travagliad:
## Problem

Linode resources created for ephemeral QA runs were not fully cleaned up on teardown, and billed for weeks:

1. **Account tags** — tags are account-level objects. Creating a VM/cluster auto-creates its tags, but destroying it leaves them behind. Every run leaked its unique `pmm-qa-run:<id>` / `expires-<epoch>` tag.
2. **LKE Block Storage volumes + NodeBalancers** — `lke cluster-delete` removes the cluster and its nodes but does **not** cascade to the CSI-provisioned volumes (`pvc-*`) or the CCM-provisioned NodeBalancer; the in-cluster controllers that would delete them die with the cluster. This was the dominant leak — an August invoice audit found **$491/mo** of orphaned volumes + NodeBalancers from HA clusters that had lived only minutes.

## Changes

**Tag cleanup**
- `prune-tags.sh` — deletes orphan tags, verifying each is attached to nothing via `GET /tags/<label>` (page_size 25) immediately before deleting; scoped to `pmm-qa*` / `expires-`; curl timeouts; non-zero on failure.
- `down.sh`, `destroy-lke.sh`, the VM self-destruct, and the LKE reaper delete each run's own tags on teardown (self-destruct deletes the instance before its tag, so a failed delete keeps the handle).

**Volume + NodeBalancer cleanup (the big leak)** — by **positive attribution**, never "unattached = orphaned":
- `create-lke-pmm-ha.sh` stamps each cluster's volumes with its `pmm-qa-run:<id>` tag at provision time (best-effort).
- `prune-lke-orphans.sh` deletes a volume **only** when it carries a `pmm-qa-run:<id>` tag whose run has **no live cluster**, and a NodeBalancer **only** when its `lke<id>` cluster is gone. A live cluster's volumes (unattached during provisioning/failover) and other owners' `pvc-*` are never in range. Fail-safe: a missed tag = a leak (recoverable), never a wrong delete.
- The **relay reaper** runs the sweep each tick (a just-deleted cluster's volumes detach slightly later). `destroy-lke` does **not** sweep — `cluster-delete` is async, so an immediate sweep would only see other runs' resources.

**Token scope** — the relay's `LINODE_TOKEN` now needs **Volumes: Read/Write** and **NodeBalancers: Read/Write** in addition to LKE R/W (documented in `SKILL.md`). Without them the sweep 401s — this was itself part of the original leak.

## Validation

- `shellcheck -S warning` clean on all scripts; `node --check` on `relay.js`.
- The pre-existing account backlog of orphaned volumes + NodeBalancers was cleared manually (dry-run reviewed first); post-cleanup, all remaining volumes were attached to live clusters.


travagliad:
Thanks — the review caught two real bugs, both fixed in `6e5d9845`:

**Account-global tag deletion could strip a live resource's tag.** `DELETE /v4/tags/<label>` removes the tag from *every* resource carrying it, and LKE run/`expires-` tags aren't unique, so the reaper's hand-rolled per-cluster deletion (and prune's incomplete instances+clusters-only inventory, which ignored Volumes/NodeBalancers/Domains) could strip a tag off a still-live cluster — losing `expires-` would make it outlive its TTL. Fixed by rewriting `prune-tags.sh` to verify each candidate is orphan via `GET /tags/<label>` (which lists tagged objects of **all** types) immediately before deleting, and having the reaper call `prune-tags.sh` instead of deleting inline. This also covers the TOCTOU recheck concern (the check is right before the delete).

Also addressed:
- **`--all` dropped** — it removed the prefix bound on a shared account; the historical bare-`nightly` sweep is a one-time by-hand job and this PR stops producing bare tags.
- **curl timeouts** (`--connect-timeout`/`--max-time`) on every API call, so a stall can't hang a teardown.
- **Non-zero exit** when any delete fails, so callers can tell orphans remain.
- **VM self-destruct** now deletes the instance before its tag (a failed delete keeps the only handle).
- **destroy-lke.sh** logs a visible message when `prune-tags.sh` isn't found instead of skipping silently.
- Doc wording fixed ("matching orphan tags"), and the duplicated comments trimmed.

`prune-tags.sh` passes `shellcheck -S warning`.

---

travagliad:
Thanks — the 🔴 was a real data-loss bug and you're right. Reworked in `104babd5` + `0cd904d0`:

**Positive attribution instead of "unattached = orphaned".** `create-lke` now stamps every one of a cluster's Block Storage volumes with its `pmm-qa-run:<id>` tag (best-effort, post-rollout). `prune-lke-orphans.sh` deletes a volume **only** when it carries a `pmm-qa-run:<id>` tag whose run has **no live cluster** — so a live cluster's volumes (unattached during Immediate-binding provisioning or failover) and any untagged/other-owner `pvc-*` are never in range. NodeBalancers stay gated on `lke<id>` cluster-gone. Fail-safe: a missed tag means a leak (recoverable), never a wrong delete.

**Dropped the `destroy-lke` sweep.** As you noted, `cluster-delete` is async, so an immediate sweep would only ever see *other* runs' resources. The reaper (which runs the sweep every tick) owns it now.

**CodeRabbit fixes:**
- Ownership before deletion — same positive-attribution change above.
- `prune-tags.sh` `GET /tags/<label>` used `page_size=1`; Linode's min is 25, so under `curl -f` it could fail and skip a real orphan-tag delete → now `page_size=25`.
- SKILL.md now gives `prune-lke-orphans.sh`'s real path (`.claude/skills/linode-ha-provisioning/scripts/…`), not "same dir".

**Comments** trimmed to drop restated script-header text.

On the SkillSpector `PE3 credential access` / `E1 external transmission` flags against SKILL.md: those are the **documented cleanup commands** that source the relay's own `/opt/pmm-ai-relay/.env` to load `LINODE_TOKEN` and call `api.linode.com` — legitimate use of the box's own token against the Linode API, in docs, not exfiltration. No change there.

---

kiranvuyurru (approved):

travagliad:
Good catches — all four are correct.

**Fixed now (`1b116478`):** the `prune-tags.sh` safety bug (finding on `prune-tags.sh:58`). `GET /tags/<label>` doesn't list LKE clusters, so a live cluster's tags could read as orphaned during its build window and get stripped globally (making it unreapable). The pruner now unions live clusters' tags (`GET /lke/clusters`) into a keep-set and never deletes one a live cluster carries.

**The other three** (`create-lke:256` failure/pending path, `create-lke:260` pagination, `prune-lke-orphans:70` NodeBalancer has no ownership tag) are all correct and share one root cause: **post-hoc tagging is fragile.** The right fix is tag-at-creation — a StorageClass carrying `linodebs.csi.linode.com/volumeTags: pmm-qa-ephemeral,pmm-qa-run:<id>` (so the CSI driver stamps every volume on every path, including failed bring-ups and still-`Pending` PVCs), plus stamping the `lke<id>-*` NodeBalancer with the same run tag and requiring it before deletion. That touches the provisioning/binding path and should be validated on a live throwaway LKE cluster before it ships.

Proposed direction, pending the maintainer: **descope the automated volume/NodeBalancer sweep to a follow-up PR** (StorageClass `volumeTags` + tagged NB, validated live), and land this PR as the solid, reviewed parts — tag cleanup (with the finding-4 fix), the VM-path teardown, the token-scope docs, and `prune-lke-orphans.sh` kept as a **hand-run** tool rather than wired into the reaper. The account backlog is already cleared, so nothing accrues meanwhile. Confirming the split with the team before regrouping the commits.

---

travagliad:
Update — went ahead and fixed all three of the remaining findings without the invasive StorageClass change, since two of them (failed bring-ups; deleting another owner's NodeBalancer) matter regardless of the pagination edge (`8749e2cb`):

- **Failed/timed-out bring-ups (`create-lke:256`):** tagging now runs in the **EXIT trap** (`_tag_for_teardown`), so a cluster that dies at a rollout/`readyz` gate still gets its volumes + NodeBalancer tagged — exactly the runs that used to leak.
- **Pagination (`create-lke:260`):** list with `--page-size 500` (and list once, not per node).
- **NodeBalancer had no ownership (`prune-lke-orphans:70`):** `create-lke` now stamps the cluster's `lke<id>-*` NodeBalancer with `pmm-qa-run:<id>`, and the sweep **requires that tag** before deleting — so another owner's NB, or a deliberately-kept migrate target, is never removed just because its cluster id is gone. Same fail-safe positive attribution as volumes.

Residual, accepted: a volume that was created but never attached at the instant a *failed* run exits won't be caught by node-attachment tagging (only a StorageClass `volumeTags` would). That's a rare leak (recoverable by hand), never a wrong delete — the true tag-at-creation version is the natural follow-up if it ever shows up in billing.

`shellcheck -S warning` clean; tag-extraction logic verified.

---

PR #1324, opened 2026-09-04
Add Performance testing scripts

shrutipradhan-percona:


travagliad (approved):

travagliad:
Closing this PR — the performance scripts have been consolidated into a security-hardened version opened against `main` as #1333 (adds env-based secrets instead of the hardcoded root password/SSH key, a teardown script so the VMs stop billing, tag-scoped batch identity, and the robustness fixes from the reviews here).

@shrutipradhan-percona — nothing of yours is deleted: your `performance-scripts` branch and its commits remain on your fork, so your local workflow still runs exactly as it does today. If you'd rather iterate here instead, just reopen. One security follow-up regardless of which PR lands: the root password and SSH key in this branch's history are on a public repo and should be rotated and scrubbed.

---

PR #1332, opened 2026-09-04
performance: harden Linode load-test scripts + add teardown

travagliad:
## What this is

A `performance/` load-test harness for PMM clients on Linode: provision many client VMs from a Percona StackScript, build an Ansible inventory of one batch, and upgrade the PMM client across it. Used for scale/load testing; not wired into CI.

It is a **hardened alternative to #1324** ("Add Performance testing scripts") — same intent, with the security and cost-leak problems fixed. Approve/merge this and close #1324, or cherry-pick from it. cc @shrutipradhan-percona, whose scripts this reworks. (Against `main` every file here is new; the comparison to #1324 is context, not part of the diff.)

## What the scripts do

- **`linode_stackScript_load_pmm3.sh`** — creates N client VMs for one DB type (mysql/postgresql/mongodb). Root password and authorized SSH key come from the environment (`LINODE_ROOT_PASS`, `PMM_PERF_SSH_PUBKEY`), never from source; the PMM admin password comes from `PMM_SERVER_PASSWORD`. `--stackscript_data` is built with `jq` so values can't break the JSON. Each VM is labelled and tagged with a `PERF_RUN_ID` (`pmm-qa-perf-run:<id>`) so a batch is uniquely identifiable and two runs can coexist.
- **`teardown_perf_linodes.sh`** — deletes the VMs (they bill until removed). Scoped to this harness's own `pmm-qa-perf` tag — never the account-wide `pmm-qa-ephemeral`, which the Terraform QA runner and LKE clusters also carry — and matches the tag as an exact array element. `--dry-run` previews; `PMM_PERF_TAG=pmm-qa-perf-run:<id>` targets one batch.
- **`prepare_ansible_client_inventory.sh`** — writes the inventory for one batch, selected by its `pmm-qa-perf-run:<id>` tag; exits non-zero if no host matches (so a mistyped batch can't report success over nothing). The generated inventory is git-ignored.
- **`start_upgrade_v3.sh` / `upgrade_client.yml` / `upgrade_all_v3.sh`** — upgrade the PMM client in every `*client_container*` Docker container to a requested version/repo. SSH host-key checking is on by default; playbook vars are shell-quoted; the requested version is pinned (fails fast if that build is unavailable); the agent is restarted with `setsid`; the upgrade runs as a monitored async job that fails on non-zero exit.
- **`README.md`** documents the full provision → inventory → upgrade → teardown workflow and the required env vars.

## Validation

- `bash -n` and `shellcheck -S warning` clean on all scripts; `yamllint` clean on the playbook; the tag exact-match and batch-selection logic checked against sample data.
- **Not run end-to-end** — provisioning needs live Linode, Ansible, and Docker client containers not available here, so the robustness claims rest on the static reading. A live smoke test before merge is worth doing.

## Follow-up outside this PR

The root password and personal SSH key committed in #1324 already exist in that PR's git history on a public repo. Removing them here does not undo that — they should be **rotated and scrubbed from history** separately.

https://claude.ai/code/session_01DBhvZuEUQU8ym347Xat4Yy


travagliad:
Addressed all review findings across two pushes (`a0148b5e`, `a238118b`):

**🔴 teardown default tag** — `teardown_perf_linodes.sh` now defaults to this harness's own `pmm-qa-perf` tag (never the account-wide `pmm-qa-ephemeral` the TF runner and LKE clusters also carry), and matches the tag as an exact array element via `linode-cli --json | jq index()` (no more `grep -qw` substring match). Verified: the exact-match excludes both a shared-tag QA-runner box and a `pmm-qa-perf-run:<id>` substring.

**🟡 batch identity** — `PERF_RUN_ID` is now in the create label (so two runs of the same dbtype/metrics_mode don't collide on Linode's unique-label rule) and the inventory selects a batch by its `pmm-qa-perf-run:<id>` tag instead of the account-wide `sp_fb_` label.

**🟡 inventory zero-match** — exits 1 when no host matches, so a mistyped batch can't report a successful upgrade of nothing.

**🟡 unreachable version fallback** — `upgrade_all_v3.sh` now fails fast when the pinned `pmm-client=<version>*` is unavailable, instead of an `else` branch that could never verify green.

**🔵 argv exposure** — softened the usage note: the PMM password still reaches `linode-cli` via `--stackscript_data` (visible in `ps`).

**🟡 comment verbosity** (create/teardown/inventory/start_upgrade/upgrade_all/playbook headers) — trimmed the change-narrating comments per the repo's minimal-comment house style; kept the `setsid` and tag-scope invariants and the README usage.

`bash -n` + `shellcheck -S warning` clean on all scripts, `yamllint` clean on the playbook. Still statically validated only — a live smoke test before merge is worth doing.

---

travagliad:
Addressed the CodeRabbit review in `92cf9261` (it reviewed `693b0dab`, before the two fix commits — three of its findings, the label-prefix grep, tag word-match, and version fallback, were already fixed there and it auto-marked them ✅).

Remaining findings, now fixed:
- **Reject unknown/extra args (teardown, 🟠)** — strict `case`-based parse; `--dry-rnu` or any extra arg hard-errors instead of deleting. Verified.
- **`bash -c` injection (upgrade_all, 🟠 CWE-78)** — now `docker exec --env DEBIAN_FRONTEND=noninteractive "$c" apt-get install -y "pmm-client=${version}*"`; no second shell parse of `${version}`.
- **Quote key path in inventory (🟡)** — the INI value is now double-quoted so a path with spaces isn't split.
- **Inventory ignore contract (🟡 CWE-200)** — dropped the `INVENTORY_FILE` override; the scripts write only the git-ignored `inventory_client_container2`.
- **Host-key enrollment (🟠)** — documented the `known_hosts` precondition in the README (kept `ANSIBLE_HOST_KEY_CHECKING=True` default; no blind `ssh-keyscan`).

One thing I could not verify without a live container: whether apt honors the `pmm-client=<version>*` version spec. If it doesn't, the pinned install fails loudly (the script exits non-zero) rather than installing the wrong version — worth confirming in the smoke test.

`shellcheck -S warning` + `bash -n` + `yamllint` clean.

---

travagliad:
Closing this in favor of directing these changes at #1324 instead of `main`. The same commits, rebased onto @shrutipradhan-percona's `performance-scripts` branch (so the diff is only the hardening), are on branch `claude/pmm-qa-hooks-permissions-verify-pt9u2v`.

Open the PR into #1324's branch here (one click): https://github.com/shrutipradhan-percona/pmm-qa/compare/performance-scripts...percona:claude/pmm-qa-hooks-permissions-verify-pt9u2v

(I couldn't open it automatically — this session can't act on the fork.)

---

travagliad:
Reopened against `main`. The cross-fork route into #1324's branch couldn't be merged, so this stays the single hardened PR to `main`, superseding #1324 (now closed). Head is back on the clean main-based commit — the diff is just the `performance/` scripts.

---

travagliad:
Correction: GitHub won't reopen this PR because the branch was force-pushed after it closed. The change is reopened as a fresh PR to `main` instead: #1333. Please follow #1333.

---

PR #1333, opened 2026-09-04
Harden performance load-test scripts + add teardown

travagliad:
## What this is

A `performance/` load-test harness for PMM clients on Linode: provision many client VMs from a Percona StackScript, build an Ansible inventory of one batch, and upgrade the PMM client across it. Used for scale/load testing; not wired into CI.

Supersedes #1324 (closed) — same intent, with the security and cost-leak problems fixed. Reopens the change that was previously on #1332 (that PR couldn't be reopened after a branch force-push). cc @shrutipradhan-percona, whose scripts this reworks; her original branch is preserved on her fork.

## What the scripts do

- **`linode_stackScript_load_pmm3.sh`** — creates N client VMs for one DB type (mysql/postgresql/mongodb). Root password and authorized SSH key come from the environment (`LINODE_ROOT_PASS`, `PMM_PERF_SSH_PUBKEY`), never from source; the PMM admin password from `PMM_SERVER_PASSWORD`. `--stackscript_data` built with `jq`. Each VM labelled and tagged with a `PERF_RUN_ID` (`pmm-qa-perf-run:<id>`) so a batch is uniquely identifiable and two runs coexist.
- **`teardown_perf_linodes.sh`** — deletes the VMs (they bill until removed). Scoped to this harness's own `pmm-qa-perf` tag (exact array match), never the account-wide `pmm-qa-ephemeral`; strict arg parse; `--dry-run` previews.
- **`prepare_ansible_client_inventory.sh`** — writes the inventory for one batch, selected by its `pmm-qa-perf-run:<id>` tag; exits non-zero if no host matches. Generated inventory is git-ignored.
- **`start_upgrade_v3.sh` / `upgrade_client.yml` / `upgrade_all_v3.sh`** — upgrade the PMM client in every `*client_container*` Docker container. SSH host-key checking on by default (README documents the `known_hosts` precondition); playbook vars shell-quoted; `apt-get` via `docker exec --env` + argv (no shell re-parse of `version`); version pinned with fail-fast; `pmm-agent` restarted with `setsid`; upgrade runs as a monitored async job that fails on non-zero exit.
- **`README.md`** documents provision → inventory → upgrade → teardown and the required env vars.

## Validation

`bash -n` + `shellcheck -S warning` clean on all scripts; `yamllint` clean on the playbook; tag exact-match and batch-selection logic checked against sample data. **Not run end-to-end** (needs live Linode/Ansible/Docker), so a smoke test before merge is worth doing — in particular confirm apt honors the `pmm-client=<version>*` spec.

## Follow-up outside this PR

The root password and personal SSH key committed in #1324 exist in that branch's history on a public repo. Removing them here does not undo that — they should be **rotated and scrubbed from history** separately.

https://claude.ai/code/session_01DBhvZuEUQU8ym347Xat4Yy

---
_Generated by [Claude Code](https://claude.ai/code/session_01DBhvZuEUQU8ym347Xat4Yy)_

### DIFF
--- a/performance/linode_stackScript_load_pmm3.sh
+++ b/performance/linode_stackScript_load_pmm3.sh
@@ -1,12 +1,7 @@
 #!/usr/bin/env bash
 # Provision N Linode PMM client load-test VMs for one DB type via a Percona
-# StackScript. Every instance is tagged pmm-qa-ephemeral so its teardown
-# counterpart (teardown_perf_linodes.sh) can find and delete it -- these VMs bill
-# until removed, so nothing here creates an untagged, unattributable instance.
-#
-# Credentials come from the environment, never from source: the VM root password
-# and the authorized SSH key were previously hardcoded here and must stay out of
-# the repo.
+# StackScript. Each instance is tagged pmm-qa-perf-run:<PERF_RUN_ID> so its batch
+# can be found for inventory and teardown; credentials come from the environment.
 set -Eeuo pipefail
 
 usage() {
@@ -17,12 +12,14 @@ Usage: linode_stackScript_load_pmm3.sh <pmm_server_host> <client_version> <insta
 Required environment:
   LINODE_ROOT_PASS     root password for the provisioned VMs
   PMM_PERF_SSH_PUBKEY  SSH public key to authorize on the provisioned VMs
-  PMM_SERVER_PASSWORD  PMM server admin password (kept off this script's argv)
+  PMM_SERVER_PASSWORD  PMM server admin password (from env; still reaches
+                       linode-cli via --stackscript_data, so visible in ps)
 
 Optional environment:
   LINODE_REGION        default: us-east
   LINODE_TYPE          default: g6-standard-2
-  PERF_RUN_ID          groups this batch under tag pmm-qa-perf-run:<id>; default: UTC timestamp
+  PERF_RUN_ID          batch id, tagged pmm-qa-perf-run:<id> and put in the
+                       label so batches don't collide; default: UTC timestamp
 EOF
   exit 2
 }
@@ -54,8 +51,7 @@ command -v jq >/dev/null || { echo "jq is required" >&2; exit 1; }
 
 for i in $(seq 1 "$INSTANCES"); do
   hostname="${hostprefix}_${METRICS_MODE}_${i}"
-  label="sp_fb_${i}_${DBTYPE}_${METRICS_MODE}_test"
-  # jq builds the JSON so a value containing a quote or backslash can't break it.
+  label="sp_fb_${i}_${DBTYPE}_${METRICS_MODE}_${PERF_RUN_ID}"
   ssdata="$(jq -nc \
     --arg hostname "$hostname" \
     --arg pmmserver "$PMM_SERVER_HOST" \
@@ -79,5 +75,6 @@ for i in $(seq 1 "$INSTANCES"); do
   sleep 15
 done
 
-echo "created ${INSTANCES} ${DBTYPE} client(s), tagged pmm-qa-ephemeral / pmm-qa-perf-run:${PERF_RUN_ID}"
-echo "tear them down with: ./teardown_perf_linodes.sh    (or --dry-run to preview)"
+echo "created ${INSTANCES} ${DBTYPE} client(s) as batch PERF_RUN_ID=${PERF_RUN_ID}"
+echo "inventory:  PERF_RUN_ID=${PERF_RUN_ID} ./prepare_ansible_client_inventory.sh"
+echo "teardown:   PMM_PERF_TAG=pmm-qa-perf-run:${PERF_RUN_ID} ./teardown_perf_linodes.sh"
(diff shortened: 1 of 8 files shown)


travagliad:
Addressed all 11 findings (commits `0e36289b`, `27…` head):

**Cost-leak / safety**
- **Create leak & misleading tag** — dropped `--tags pmm-qa-ephemeral` (nothing reaps stray Linode instances by it); the `PERF_RUN_ID` + teardown command now print **before** the loop, and an `ERR` trap tears the batch down on any mid-loop failure, so a partial run can't leak billing VMs.
- **Unpaginated listing** — `teardown` and `prepare_…` now page `linode-cli linodes list` at `--page-size 500` until a short page, so a busy account can't hide instances behind the default 100 rows.
- **`PMM_PERF_TAG` could point at a shared tag** — now validated to `pmm-qa-perf` or `pmm-qa-perf-run:<id>` only (hard-error otherwise), and documented in the README env table.
- **Delete stdin** — the `linode-cli delete` reads from `/dev/null` so it can't consume the row here-string and exit the loop early with a false `failed=0`.

**Correctness**
- **apt version glob** — replaced `pmm-client=<version>*` (apt doesn't glob the version half) with resolving the concrete published build via `apt-cache madison` and installing that exact version; still fails loudly if none matches.
- **Unknown args in `upgrade_all_v3.sh`** — now a hard error (matched to the teardown parser) instead of silently ignored.
- **Unused `extra_repo`/`extra_version` layer** — collapsed to the `lookup('env', …)` form actually exercised.

**Comments** — trimmed the change-narrating ones; added the one that was missing (why the `sleep 15` exists).

One I resolved differently: the `inventory_client_container2` **sync-comment** — I dropped the comment but kept the name hardcoded rather than reintroducing a configurable `PMM_PERF_INVENTORY`, because a configurable inventory path was flagged as a disclosure risk (CWE-200) in the earlier CodeRabbit review. Happy to add a scoped override if you'd prefer.

`shellcheck -S warning` + `bash -n` clean; tag-validation, pagination-break, and madison-resolution logic checked against sample data. Still not run end-to-end.

---

performance/linode_stackScript_load_pmm3.sh
```diff
@@ -0,0 +1,92 @@
...
+echo "inventory:  PERF_RUN_ID=${PERF_RUN_ID} ./prepare_ansible_client_inventory.sh"
+echo "teardown:   PMM_PERF_TAG=pmm-qa-perf-run:${PERF_RUN_ID} ./teardown_perf_linodes.sh"
+
+# Nothing reaps stray Linode instances automatically, so a mid-loop failure would
+# leak billing VMs -- tear this batch down on any error instead.
+cleanup_on_err() {
+  echo "provisioning failed -- tearing down batch pmm-qa-perf-run:${PERF_RUN_ID}" >&2
+  PMM_PERF_TAG="pmm-qa-perf-run:${PERF_RUN_ID}" bash "$(dirname "$0")/teardown_perf_linodes.sh" || true
+}
+trap cleanup_on_err ERR
```
kiranvuyurru: job cancellation, or timeout can still leave billable VMs behind.

travagliad: Good catch — fixed in two layers, since a bash `ERR` trap doesn't fire on the SIGTERM/SIGINT a CI cancel or timeout sends:

1. **This script** now also traps `TERM`/`INT`/`HUP` and runs the same batch teardown, so a cancelled/timed-out run cleans up after itself (best-effort).
2. **The pipeline** (draft #1334) tears down on `success() || cancelled()`, so a cancel/timeout at the job level also cleans up, while a genuine test failure is still kept for investigation.

Neither can survive a hard `SIGKILL` or a runner dying mid-run, so the real guarantee is an **out-of-band tag-expiry reaper** (an `expires-<epoch>` tag + a scheduled sweep of `pmm-qa-perf` instances) — noted as a TODO on #1334. Want me to add that reaper too?

---

kiranvuyurru (approved):

travagliad:
Addressed the latest CodeRabbit review (pushed):

- **🟠 Upgrade loop aborting** — each container's upgrade now runs in a subshell, so one failure is recorded in `rc` and the loop continues to the rest instead of `set -e` aborting the whole run.
- **🟡 Inventory destroyed on no-match** — `prepare_ansible_client_inventory.sh` builds into a temp file and validates the host count before `mv` into place, so a zero-match run leaves an existing inventory untouched.
- **🟡 Literal version match** — the madison lookup uses `index($2, v)==1` and verification uses `grep -Fq`, so a value like `.*` can't select an arbitrary build or pass verification.
- **🟠 CWE-214 (creds on argv)** — `LINODE_ROOT_PASS` is now generated random per run when unset (access is by SSH key, so it's throwaway) → what reaches `--root_pass` is a single-use secret for a VM that gets destroyed. `PMM_SERVER_PASSWORD` in `--stackscript_data` is inherent to `linode-cli` (the API needs it to register clients) and can't be kept off argv without a different provisioning tool; mitigated by running on single-tenant ephemeral CI runners. Flagged as the residual.

`shellcheck -S warning` + `bash -n` clean; the literal-match and subshell logic checked against samples.

---

PR #1366, opened 2026-09-08
Make the pmm-client download survive a slow repo.percona.com

travagliad:

- source: https://github.com/percona/pmm-qa/actions/runs/34233153827
- tests:
  - `.github/workflows/runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` / setup shard `ps-replication`, step **Setup PMM-Client against remote PMM Server** (no test spec — the failure is in setup, and it took the whole run's tests with it)

## What happened

Nightly run **316** (`SERVER_IP=3.141.43.210`, `client_version=3.8.1`) failed on exactly one job: `setup / ps-replication`. The step runs

```
sudo timeout -k 30 9m bash -x pmm3-client-setup.sh --client_version 3.8.1 …
```

which lands in the released-version branch of `pmm3-client-setup.sh` and did a bare
`wget -O pmm-client.deb https://repo.percona.com/…/pmm-client_3.8.1-1.jammy_amd64.deb` —
`Length: 178025706 (170M)`.

From the job log:

| | reached | rate | outcome |
|---|---|---|---|
| Attempt 1 (13:50:51 → 13:59:29) | `94650K` — **54%** | ~190 KB/s | `Code: 124` |
| Attempt 2 (13:59:43 → 14:08:40) | `113750K` — **65%** | ~250 KB/s | `Code: 124`, `Final attempt failed` |

`--continue` was absent, so attempt 2 threw away attempt 1's 94 MB and re-raced the same
9-minute clock from byte 0. Then `fail-fast: true` on the `setup` matrix cancelled the other
13 setup shards, and all three `test execution / …` jobs failed in **Wait for all setup jobs
to be ready** having executed **zero tests**.

## Not the product, and not a broken artifact

Re-checked the URL during triage: `HTTP/2 200`, `content-length: 178025706`,
`etag: "6a31348d-a9c74ea"`, `last-modified: Tue, 16 Jun 2026 11:33:33 GMT` — unchanged and
healthy, and it served the full 170 MB in 11 s (16 MB/s) from the triage host. The origin is
fine; the runner→OVH path was slow at that moment. Throughput for this same file *within the
same run* spanned **0.19–3.8 MB/s**:

| shard | step 7 duration |
|---|---|
| `mysql` | 1m15s |
| `pgsql` | 2m16s |
| `pxc` | **8m18s** — under the 9m wall by 42 s |
| `ps-replication` | never finished |

So `pxc` was one bad minute away from failing the same way. This is not specific to
`ps-replication`: today's other nightly failures tripped on a different shard each time
(`setup / ps-gr` in run 314, `setup / extra-pxc` in run 312).

## The fix

Every place that fetches the PMM client for CI. Review of the first commit established that
hardening only `pmm3-client-setup.sh` left the same defect on the path that installs the
client *inside each monitored container* — see the resolved threads.

**Download hardening** — `--continue` where the URL is immutable, `--timeout=60
--waitretry=15` everywhere, `--progress=dot:giga` so a failing step stays readable (the
default dot style emitted ~2000 progress lines per attempt, which is what made the original
failure unreadable from a log tail):

| file | site | resumable? |
|---|---|---|
| `pmm3-client-setup.sh` | deb (released version) | yes — partial named after the version |
| `pmm3-client-setup.sh` | tarball | **no** |
| `pmm3-client-setup-centos.sh` | rpm (released version) | yes — partial named after the version |
| `pmm3-client-setup-centos.sh` | tarball | **no** |
| `tasks/install_pmm_client.yml:151` | deb, in-container | yes — `rm -f` first |
| `tasks/install_pmm_client.yml:170` | rpm, in-container | yes — `rm -f` first |
| `tasks/install_pmm_client.yml:183` | tarball, in-container | **no** |

The tarball paths deliberately keep restarting: `pmm-client-latest.tar.gz` is mutable under a
fixed name, so resuming into it would splice two different builds. The two scripts run in the
runner workspace, which survives an outer retry, so their partials are version-named and
resume *across* attempts. The container paths cannot — `on_retry_command` wipes every
container — so there the deb and rpm are removed first and `--continue` serves only wget's own
internal retries, with no stale file to splice into.

`tasks/install_pmm_client.yml` matters because ten setup playbooks include it (mysql,
percona-server, postgresql, pdpgsql single/replication/patroni, valkey sentinel/cluster,
external, add_mysql), making it what runs during **Run setup for E2E tests** — the sibling
step in these same two workflows, under the same budget.

**No `--tries`.** An earlier revision set `--tries=3`; that replaced wget's default of **20**
and did not do what it was added for. A stall is bounded by `--timeout=60`; `--tries` caps how
many times wget *recovers*, and with `--continue` each recovery is a resume. Three connection
resets and wget would exit, discarding the rest of the budget this PR buys. The outer
`timeout -k 30 19m` is the bound.

**Budget: 9m → 19m** (`timeout_minutes: 10 → 20`) in the two workflows that wrap
`pmm3-client-setup.sh` in a clock — `runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` and
`ha-e2e-tests.yml`. This matches the sibling **Run setup for E2E tests** step already at
20m/19m in both files, and sits well inside the job budgets (120 min nightly, 90 min HA).
`ha-e2e-tests.yml` carried the byte-identical 9m wrapper, so it had the same latent failure.

## Verification

Against the real artifact URL, not a mock.

**1. Resume produces a byte-identical file.** Killed a download at 135,757,469 bytes, re-ran
with `--continue`: wget skipped the existing data and finished at exactly `178025706`.
`md5 = 00e2a5ae3b7284663c4fae51b090a5bf`, identical to a full fresh `curl` of the same URL.

**2. The failure reproduces under the wrapper, and the fix changes the outcome.** Two
attempts of a 20 s `timeout` against a `--limit-rate=4m` cap — the same shape as CI's
`2 × timeout 9m` at ~190 KB/s:

```
OLD (no --continue)  attempt 1: exit=124 size=81641117   attempt 2: exit=124 size=82099869
NEW (--continue)     attempt 1: exit=124 size=75087517   attempt 2: exit=124 size=156761360
```

The old command's second attempt lands where the first did — the real run's 54% → 65% dead
end. The new one carries 75 MB forward and reaches 88% of the file on the same total budget;
at 2 × 19m rather than 2 × 20s it completes with wide margin.

**Static:** `.claude/hooks/lint-changed.sh` (yamllint + shellcheck) clean on all changed
files; `bash -n` clean on both scripts; the ansible task file parses.

### Reproduction gap, stated plainly

**No Linode VM was provisioned.** The failure is a `timeout` killing a `wget` on a GitHub
runner, and the log carries the byte counts, rates and both exit-124s directly — and a
throwaway VM on a fast link cannot reproduce a runner→OVH network path either way. The
mechanism was instead reproduced against the real URL under the same wrapper shape, above.
Nightly's PMM Server is external and already gone, so the original run cannot be re-run.

**What only a real run can confirm:** that a nightly at ~190 KB/s now completes. A fresh
nightly against this branch is available if wanted — the dispatching Jenkins job takes
`PMM_QA_GIT_BRANCH`.

## Left out deliberately: `fail-fast: true`

The download timeout is the *cause*; `fail-fast: true` on the `setup` matrix
(`nightly-e2e-tests-matrix.yml:80`) is the *amplifier* that turned one slow shard into 13
cancelled shards and a nightly that reported nothing. Flipping it to `false` is not a
one-liner: the test-execution runners wait on `expected_setup_jobs: 14`, so surviving shards
would need that counting logic reworked to distinguish "failed" from "not ready yet", plus its
own verification. Worth doing as its own change — flagging it rather than smuggling it in here.

https://claude.ai/code/session_01S56rE1VTWvwwNpa5CBgy1e

---
_Generated by [Claude Code](https://claude.ai/code/session_01S56rE1VTWvwwNpa5CBgy1e)_

### DIFF
--- a/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
@@ -90,8 +90,11 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
   elif [ "$client_version" = "3.8.1" ] || [ "$minor_version" -gt 8 ]; then
     build_number=1
   fi
-  wget -O pmm-client.rpm https://repo.percona.com/pmm3-client/yum/release/9/RPMS/x86_64/pmm-client-${client_version}-${build_number}.el9.x86_64.rpm
-  rpm -i pmm-client.rpm
+  rpm_file="pmm-client-${client_version}-${build_number}.el9.x86_64.rpm"
+  # Resumable: a released version's URL is immutable and the file name carries the version.
+  wget --continue --timeout=60 --waitretry=15 --progress=dot:giga \
+    -O "${rpm_file}" "https://repo.percona.com/pmm3-client/yum/release/9/RPMS/x86_64/${rpm_file}"
+  rpm -i "${rpm_file}"
 fi
 
 ## Default Binary path
@@ -102,7 +105,10 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       wget -O pmm-client.tar.gz --progress=dot:giga "${client_version}"
+       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
+       # left by an earlier build must not be resumed into.
+       wget -O pmm-client.tar.gz --progress=dot:giga \
+         --timeout=60 --waitretry=15 "${client_version}"
     fi
     tar -zxpf pmm-client.tar.gz
     rm -r pmm-client.tar.gz
--- a/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
@@ -105,8 +105,6 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
-       # left by an earlier build must not be resumed into.
        wget -O pmm-client.tar.gz --progress=dot:giga \
          --timeout=60 --waitretry=15 "${client_version}"
     fi
--- a/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup-centos.sh
@@ -91,7 +91,6 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
     build_number=1
   fi
   rpm_file="pmm-client-${client_version}-${build_number}.el9.x86_64.rpm"
-  # Resumable: a released version's URL is immutable and the file name carries the version.
   wget --continue --timeout=60 --waitretry=15 --progress=dot:giga \
     -O "${rpm_file}" "https://repo.percona.com/pmm3-client/yum/release/9/RPMS/x86_64/${rpm_file}"
   rpm -i "${rpm_file}"
--- a/qa-integration/pmm_qa/pmm3-client-setup.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup.sh
@@ -105,8 +105,15 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
   elif [ "$client_version" = "3.8.1" ] || [ "$minor_version" -gt 8 ]; then
     build_number=1
   fi
-  wget -O pmm-client.deb "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
-  dpkg -i pmm-client.deb
+  deb_file="pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
+  # ~170MB, and repo.percona.com has served it to CI runners as slowly as 190KB/s.
+  # Resume instead of restarting: the callers wrap this script in a retry that gives
+  # every attempt the same wall clock, so without --continue a slow link never finishes.
+  # Safe to resume here (a released version's URL is immutable, and the file name
+  # carries the version, so a stale partial can never belong to a different build).
+  wget --continue --tries=3 --timeout=60 --waitretry=15 --progress=dot:giga \
+    -O "${deb_file}" "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/${deb_file}"
+  dpkg -i "${deb_file}"
 fi
 
 ## Default Binary path
@@ -117,7 +124,10 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       wget -O pmm-client.tar.gz --progress=dot:giga "${client_version}"
+       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
+       # left by an earlier build must not be resumed into.
+       wget -O pmm-client.tar.gz --progress=dot:giga \
+         --tries=3 --timeout=60 --waitretry=15 "${client_version}"
     fi
     tar -zxpf pmm-client.tar.gz
     rm -r pmm-client.tar.gz
--- a/qa-integration/pmm_qa/pmm3-client-setup.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup.sh
@@ -106,7 +106,6 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
     build_number=1
   fi
   deb_file="pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
-  # Resumable: a released version's URL is immutable and the file name carries the version.
   wget --continue --timeout=60 --waitretry=15 --progress=dot:giga \
     -O "${deb_file}" "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/${deb_file}"
   dpkg -i "${deb_file}"
@@ -120,8 +119,6 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
-       # left by an earlier build must not be resumed into.
        wget -O pmm-client.tar.gz --progress=dot:giga \
          --timeout=60 --waitretry=15 "${client_version}"
     fi
--- a/qa-integration/pmm_qa/pmm3-client-setup.sh
+++ b/qa-integration/pmm_qa/pmm3-client-setup.sh
@@ -106,11 +106,7 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
     build_number=1
   fi
   deb_file="pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
-  # ~170MB, and repo.percona.com has served it to CI runners as slowly as 190KB/s.
-  # Resume instead of restarting: the callers wrap this script in a retry that gives
-  # every attempt the same wall clock, so without --continue a slow link never finishes.
-  # Safe to resume here (a released version's URL is immutable, and the file name
-  # carries the version, so a stale partial can never belong to a different build).
+  # Resumable: a released version's URL is immutable and the file name carries the version.
   wget --continue --tries=3 --timeout=60 --waitretry=15 --progress=dot:giga \
     -O "${deb_file}" "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/${deb_file}"
   dpkg -i "${deb_file}"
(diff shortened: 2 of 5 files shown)


qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -105,8 +105,15 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
...
     build_number=1
   fi
-  wget -O pmm-client.deb "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
-  dpkg -i pmm-client.deb
+  deb_file="pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
+  # ~170MB, and repo.percona.com has served it to CI runners as slowly as 190KB/s.
+  # Resume instead of restarting: the callers wrap this script in a retry that gives
+  # every attempt the same wall clock, so without --continue a slow link never finishes.
+  # Safe to resume here (a released version's URL is immutable, and the file name
+  # carries the version, so a stale partial can never belong to a different build).
```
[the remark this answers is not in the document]
travagliad: Agreed, and it matches this repo's own house style ("do not narrate what the code plainly says"). Applied verbatim in `4f4e5cc`; the measurements and the retry-wrapper reasoning stay in the PR body.

---

qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -105,8 +105,15 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
...
+  wget --continue --tries=3 --timeout=60 --waitretry=15 --progress=dot:giga \
```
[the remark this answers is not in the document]
travagliad: Correct, and this was a real bug in the fix rather than a style point — `--tries` bounds recoveries, not stalls, and with `--continue` each of those recoveries is a resume, so capping it at 3 threw away the budget the rest of this PR buys. Dropped in `cafcd4e`, leaving wget's default of 20 bounded by the outer `timeout -k 30 19m`.

---

qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -105,8 +105,15 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
...
+    -O "${deb_file}" "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/${deb_file}"
```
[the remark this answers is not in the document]
travagliad: Verified and fixed in `cafcd4e` — this was the most useful of the four. `tasks/install_pmm_client.yml` does re-derive the same `build_number` branch and fetch the same URL bare, at line 151 (deb) with rpm and tarball siblings at 170 and 180, and it is included by ten setup playbooks, so the fix genuinely did not reach the in-container path.

Grepping out from it also turned up `pmm3-client-setup-centos.sh` (rpm line 93, tarball 105) — the RHEL counterpart of the script this PR started with, carrying the identical defect. Both are now hardened the same way: `--continue` on the immutable deb/rpm URLs, timeout/waitretry only on the mutable tarballs. The container-side deb and rpm are `rm -f`'d first, since containers are recreated between outer retries (`on_retry_command` wipes them), so only wget's own partial can ever be resumed — which is what its internal retries need.

---

qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -117,7 +124,10 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       wget -O pmm-client.tar.gz --progress=dot:giga "${client_version}"
+       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
+       # left by an earlier build must not be resumed into.
+       wget -O pmm-client.tar.gz --progress=dot:giga \
+         --tries=3 --timeout=60 --waitretry=15 "${client_version}"
```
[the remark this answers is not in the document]
travagliad: Agreed — on a path with no `--continue`, giving up after three restarts is strictly worse than letting the default 20 run inside the `timeout`. Dropped here and in the centos script's tarball path in `cafcd4e`; `--timeout=60` and `--waitretry=15` kept.

---

qa-integration/pmm_qa/tasks/install_pmm_client.yml
```diff
@@ -148,7 +148,11 @@
       elif [ "$client_version" = "3.8.1" ] || [ "$minor_version" -gt 8 ]; then
         build_number=1
       fi
-      wget -O /pmm-client.deb "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
+      # Removed first so wget resumes only its own partial: containers are recreated
+      # between outer retries, so nothing older can be resumed into.
```
travagliad:
```suggestion
```

qa-integration/pmm_qa/pmm3-client-setup-centos.sh
```diff
@@ -102,7 +105,10 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       wget -O pmm-client.tar.gz --progress=dot:giga "${client_version}"
+       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
+       # left by an earlier build must not be resumed into.
```
travagliad:
```suggestion
```

qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -105,8 +105,11 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
   elif [ "$client_version" = "3.8.1" ] || [ "$minor_version" -gt 8 ]; then
     build_number=1
   fi
-  wget -O pmm-client.deb "https://repo.percona.com/pmm3-client/apt/pool/main/p/pmm-client/pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
-  dpkg -i pmm-client.deb
+  deb_file="pmm-client_${client_version}-${build_number}.$(lsb_release -sc)_amd64.deb"
+  # Resumable: a released version's URL is immutable and the file name carries the version.
```
travagliad:
```suggestion
```

qa-integration/pmm_qa/pmm3-client-setup.sh
```diff
@@ -117,7 +120,10 @@ ln -sf ${path}/bin/pmm-agent /usr/local/bin/pmm-agent
 
 if [[ "$client_version" == http* ]]; then
     if [[ "$install_client" == "yes" ]]; then
-       wget -O pmm-client.tar.gz --progress=dot:giga "${client_version}"
+       # No --continue: this URL is mutable (pmm-client-latest.tar.gz), so a partial
+       # left by an earlier build must not be resumed into.
```
travagliad:
```suggestion
```

qa-integration/pmm_qa/pmm3-client-setup-centos.sh
```diff
@@ -90,8 +90,11 @@ if [[ "$client_version" =~ ^3\.[0-9]+\.[0-9]+$ ]]; then
   elif [ "$client_version" = "3.8.1" ] || [ "$minor_version" -gt 8 ]; then
     build_number=1
   fi
-  wget -O pmm-client.rpm https://repo.percona.com/pmm3-client/yum/release/9/RPMS/x86_64/pmm-client-${client_version}-${build_number}.el9.x86_64.rpm
-  rpm -i pmm-client.rpm
+  rpm_file="pmm-client-${client_version}-${build_number}.el9.x86_64.rpm"
+  # Resumable: a released version's URL is immutable and the file name carries the version.
```
travagliad:
```suggestion
```

travagliad:
### CI on `cafcd4e`: 35 green, 1 skipped, 1 failure — the failure is not this PR's

**Failing check:** `FB E2E tests / PGSM UI tests / e2e tests: @pgsm-pmm-integration` ([run 34239854403](https://github.com/percona/pmm-qa/actions/runs/34239854403/job/102107380568)) — `18 passed, 1 failed, 2 skipped`.

```
PMM-T1063 - Verify Application Name with pg_stat_monitor
AssertionError: Expected only 5 Queries to show up for PMMT1063
                based on the load script but found 4
  at tests/qa-integration/pmm_pgsm_integration_test.js:418
```

**Why it isn't this PR's.** The assertion runs a load script, sleeps a fixed `I.wait(120)`, then asserts the QAN row count is *exactly* 5 over a `now-5m` window. That count is fed by an asynchronous chain — pg_stat_monitor → pmm-agent → qan-api2 → ClickHouse → UI — so 4-of-5 is one query that had not propagated inside the fixed wait, not a wrong value.

This PR cannot reach it:

- Its only behavioral change is wget flags on the pmm-client download. The installed client bytes are identical.
- This job ran with `CLIENT_VERSION: latest-tarball`, where the sole change is `--timeout=60 --waitretry=15 --progress=dot:giga` — no `--continue`, no rename.
- The step-timeout change (9m → 19m) is confined to `runner-e2e-tests-codeceptjs-remote-nightly-setup.yml` and `ha-e2e-tests.yml`. Neither is used by `e2e-tests-matrix.yml`, which is what ran here.
- The job's own client setup succeeded and 18 tests passed, several of them QAN-dependent.

**Not a standing base failure either:** `e2e-tests-matrix.yml` is green on `main` for its last 8 scheduled runs, so this reads as intermittent rather than newly broken.

**What happens next.** The two suggestion commits re-triggered the full matrix, so this job is re-running on `5d8b5c3` — that is the one re-run, and its result is the evidence. If it fails again on the same count assertion I will stop calling it intermittent and treat it as real.

**Deliberately not fixed here.** `PMM-T1063` would be made robust by polling until the expected rows land instead of a fixed sleep plus exact equality, but that is an unrelated QAN test and changing it would widen a client-download PR. Flagging it as a follow-up rather than folding it in.

---

kiranvuyurru (approved):

PR #1371, opened 2026-09-08
Accept a client tarball URL as the expected version

travagliad:

- source: https://pmm.cd.percona.com/job/pmm3-nightly-orchestrator/3/ — `pmm3-upgrade-test-runner` [#23452](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23452/), [#23453](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23453/), [#23454](https://pmm.cd.percona.com/job/pmm3-upgrade-test-runner/23454/) (upgrade from 3.7.0)
- tests: none — the runs died in setup before any test ran; this is the companion fix
- primary PR: Percona-Lab/jenkins-pipelines#4430 (**land together**)

## Why

The 3.7.0 upgrade variants failed because `pmm-client 3.7.0` is no longer in the pmm3-client apt repo, so the pool URL `pmm3-client-setup.sh` builds by hand 404s. The fix (jenkins-pipelines#4430) installs such sources from their published client tarball instead, passing `CLIENT_VERSION` as a URL — a route `pmm3-client-setup.sh` already supports and `pmm3-upgrade-test-runner.groovy` already special-cases.

`check_client_upgrade.py` does not. It compares the installed version against `arguments[1]` verbatim:

```python
if not admin_version.startswith(expected_version):
```

so a URL can never match and `Check Client before Upgrade` fails even though the client is correct. That branch of the upgrade matrix has been unreachable since it was written (`v3-old` and the iterated versions are exact complements — see #4430), which is why this has never surfaced.

## Verified

On a throwaway Linode VM against `percona/pmm-server:3.7.0`, with a real registered `pdpgsql_pgsm_ssl_17` container:

| argument | before | after |
|---|---|---|
| 3.7.0 tarball URL | exit 1 — `expected: https://…/pmm-client-3.7.0-x86_64.tar.gz actual: 3.7.0` | **exit 0** |
| `3.7.0` | exit 0 | exit 0 (unchanged) |
| `3.8.1` (genuinely wrong) | exit 1 | exit 1 (still fails, as it must) |
| URL with no version in it | exit 1 (misleading) | exit 1 — `Cannot determine expected version from client tarball URL: …` |

The wrong-version case is the one that matters: the check still rejects a mismatched client, it just compares against the version in the tarball name rather than the whole URL.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JAA4rTkQmjq8kkYvn2TWeh)_

### DIFF
--- a/support_scripts/check_client_upgrade.py
+++ b/support_scripts/check_client_upgrade.py
@@ -1,3 +1,4 @@
+import re
 import subprocess
 import sys
 
@@ -132,6 +133,14 @@ if len(errors) > 0:
   raise RuntimeError("Some errors in pmm-admin status: ".join(errors))
 
 expected_version=arguments[1].replace("\\r\\n", "").replace("-rc", "")
+
+# CLIENT_VERSION may be a client tarball URL instead of a bare version: the upgrade
+# matrix uses one for releases whose client deb is no longer in the apt repo.
+if expected_version.startswith("http"):
+  tarball_version = re.search(r"pmm-client-(\d+\.\d+\.\d+)", expected_version)
+  if not tarball_version:
+    raise RuntimeError(f"Cannot determine expected version from client tarball URL: {expected_version}")
+  expected_version = tarball_version.group(1)
 
 
 if not admin_version.startswith(expected_version):


travagliad:
The base is deliberate, not scratch — but the review is right that the body never says so, and right that the branch has no PR of its own. Both worth fixing.

`claude/nightly-orchestrator` is a staging branch created from `main` at `c4ae7ab3` as the common base for the seven PRs coming out of the `pmm3-nightly-orchestrator` #3 triage (five in Percona-Lab/jenkins-pipelines, two here). The point of a shared base is that the whole bundle can be exercised in one orchestrator run with `PMM_QA_GIT_BRANCH=claude/nightly-orchestrator` before any of it reaches `main`. `probe-upgrade.js` being its only commit so far is an accident of ordering, not evidence that the branch is a dumping ground.

Two corrections and one agreement:

- **`PMM_QA_GIT_BRANCH` defaulting to `main` is correct** and is exactly why the staging base works: the default keeps production nightlies on `main` while a validation run can be pointed at the branch explicitly.
- **The standalone claim is correct.** Only `http`-prefixed values take the new path, so with no caller passing a URL this is a no-op. Verified locally: `3.7.0` → `3.7.0`, `3.9.1-rc` → `3.9.1`, and `.../pmm-client-3.7.0-x86_64.tar.gz` → `3.7.0`. A URL with no version in its filename raises rather than silently comparing, which is the right failure.
- **There is a real ordering trap here, in the other direction.** Percona-Lab/jenkins-pipelines#4430 *requires* this change: it starts passing a tarball URL as `CLIENT_VERSION` for upgrade sources whose client deb has aged out of the apt pool. If #4430 lands while a run uses `PMM_QA_GIT_BRANCH=main` and this is still only on the staging branch, `Check Client before Upgrade` fails with a correctly installed client. Because this diff is a no-op until #4430 exists, it is the one PR in the stack that is safe — and arguably better — to land on `main` first, which removes the cross-repo ordering constraint entirely.

Retargeting is the PR author's call rather than mine, so I am not changing the base here; I have put the option to them with the reasoning above. Opening a PR for the staging branch itself is the other half and is worth doing regardless, since right now the accumulated fixes have no path to `main` and no review gate.

---

PR #1372, opened 2026-09-08
Stop two package-test rescues from hiding failures, and accept a tarball URL as the expected client version

travagliad:
The pmm-qa half of the `pmm3-nightly-orchestrator` #3 triage. Both changes were reviewed and merged onto this staging branch as #1370 and #1371; this is the single gate that carries them to `main`.

`origin/main` is merged in (it had #1369, which this branch lacked), so the three-dot diff is exactly the changes below — 6 files, +34/−5, no `pmm-framework` churn.

## 1. A diagnostic that reported itself instead of the failure (#1370)

`pkg amd64 / upgrade custom path` failed on two of eight OS branches reporting

```
open /pmm/package-testing/pmm-summary.zip: no such file or directory
```

which is not the failure. The real one is 40 lines earlier — `pmm-admin list` exiting 1 with `Service with ID "…" not found.` The outer `rescue` ran `pmm-admin summary` with **`failed_when: true`**, so the diagnostic's own failure became the play's reported error.

`/pmm/package-testing/` is never created on a package-test host: across the nine failing builds the summary was invoked 6 times and failed with `no such file or directory` **every** time. It turned out to be a rename fossil — `runner-package-test.yml:337-343` uploads `${{ github.workspace }}/pmm-summary.zip` on failure, and line 279 of the same file syncs that directory to `/pmm/pmm-qa/` inside the VM the playbook runs in. So the summary now defaults to the playbook's own repo root, which is that synced folder under Actions and the job workspace under Jenkins.

Two sibling call sites had the same shape with `ignore_errors` instead, which meant the rescue always succeeded, the block's error was cleared, and **a metric that never appeared was recorded as `rescued=1` with the play green**. Those two suites have been unable to fail on a missing metric for as long as that shape existed. All three now collect the summary without letting it decide the outcome, then `fail:` with the real error.

Seven of the ten package playbooks include one of those two files, so **suites that were passing on a missing metric will start failing**. Those are real failures that were invisible, not new ones.

## 2. A retry that was missing (#1370)

`enable_repo.yml`'s RedHat `clean all && makecache` had no `register`, no `until`, no `retries`, while the sibling apt task seven lines below has had `retries: 5` all along. `repo.percona.com` regenerates the `pmm3-client` experimental repo on the hour at ~:52, and five of the nine package failures in that run land in 13:50–14:06 with hash/size/`Content-Length` mismatches and repodata 404s — the repo changing under the client. Oracle Linux 9 died on its single attempt. It now retries like its sibling, with the two commands `&&`-chained so a failing `clean all` cannot be hidden by a succeeding `makecache` (review catch — the `shell` task returns only the last command's status).

## 3. A same-version upgrade is not an upgrade (#1370)

`pmm3-client_integration_upgrade_custom_path.yml` derived `old_version` from `latest_release.stdout`, the newest Docker Hub tag. Once Percona-Lab/jenkins-pipelines#4428 hands the arm64 tarball lanes the newest **GA** as `pmm_version`, that source becomes *the same version* — the suite would "upgrade" 3.9.1 → 3.9.1 and pass vacuously. It now steps back a release when `pmm_version` is already the newest one:

| `test_repo` | `pmm_version` | `old_version` | |
|---|---|---|---|
| experimental | 3.10.0 | 3.9.1 | amd64 today — unchanged |
| experimental | 3.9.1 | **3.9.0** | new arm64 lane — a real 3.9.0 → 3.9.1 upgrade |
| main | 3.10.0 | 3.9.0 | pre-existing behaviour preserved |
| release | 3.9.1 | 3.9.0 | |

Both `pmm-client-3.9.0-aarch64.tar.gz` and `-3.9.1-` return 200, so the new lane's two downloads exist.

## 4. `check_client_upgrade.py` accepts a tarball URL (#1371)

Required by Percona-Lab/jenkins-pipelines#4430, which installs upgrade sources whose client deb has aged out of the apt pool from their published tarball, passing `CLIENT_VERSION` as a URL. The check compared the installed version against that argument verbatim, so a URL could never match and `Check Client before Upgrade` failed with a correctly installed client. Inert until that PR is in use: only `http`-prefixed values take the new path. The URL is parsed *before* `-rc` is stripped (review catch — otherwise `…/pmm-client-3.3.0-rc1.tar.gz` captured `3.3.01`).

## Verification, and the gap that matters most

**No CI on this head executes `package_tests/**`.** All five package workflows — `package-test-fb`, `package-test-matrix`, `package-test-matrix-full`, `package-test-single`, `runner-package-test` — are `workflow_dispatch:` only, and nothing triggers on a push or pull request touching those paths. The checks that do run here are the FB E2E suites, which exercise none of the changed files. So a green PR here says nothing about the seven-playbook blast radius in §1, and the verification below *is* the evidence, not a supplement to a run.

Exercised against real `ansible-core` and the real artifacts, with the pre-change files as controls.

- **The swallowing, shown rather than argued.** A harness with the same block/rescue nesting and the summary command forced to fail: `failed=0 rescued=2 ignored=1` before, `failed=1 rescued=2 ignored=1` after, and the message that surfaces is the real one, not the summary's.
- **The masking reproduced byte-identically.** The pre-change `PLAY RECAP` matches the CI recap of both failing branches (`failed=1 … rescued=1`) along with the same misleading message.
- **The retry.** Substituting a script that 404s twice then succeeds, using the task's exact attributes: `EXIT=2 attempts=1` before, `EXIT=0 attempts=3` after — the `before` reproducing the Oracle Linux 9 failure exactly. And the `&&` chain: `false⏎true` gives rc=0, `false && true` gives rc=1.
- **The summary path.** The new default resolves to the repo root under real ansible, and an explicit `pmm_summary_file` still overrides it.
- **The URL parsing**, over every shape the matrix can produce: `3.7.0 → 3.7.0`, `3.9.1-rc → 3.9.1`, `…/pmm-client-3.7.0-x86_64.tar.gz → 3.7.0`, `…/pmm-client-3.10.0-aarch64.tar.gz → 3.10.0` (the two-digit minor matters next), `…/pmm-client-3.3.0-rc1.tar.gz → 3.3.0`, and a URL with no version raises rather than silently comparing. A genuinely wrong version still fails, which is the case that matters.
- **Static:** `yamllint --strict` clean on all changed YAML, `py_compile` clean, `ansible-playbook --syntax-check` clean on the playbooks that include the changed task files.

Worth knowing alongside that: **CI has no semantic gate for ansible either.** `.github/workflows/lint.yml` delegates to `.claude/hooks/lint-changed.sh`, which runs `yamllint --strict`, `shellcheck`, `ruff`, `eslint`/`tsc` and `terraform fmt` — no `ansible-lint`, no `--syntax-check`. Adding one is worth doing separately; it needs `ansible` in the lint job and must run only over the 10 playbooks, since the files under `tasks/` are `include_tasks` fragments that cannot be syntax-checked standalone.

**What only a real run can confirm:** that the seven playbooks including the two un-swallowed task files behave as expected, and which previously-green suites now fail on a metric that was always missing. A `workflow_dispatch` of `package-test-matrix` against this branch, or a `pmm3-nightly-orchestrator` run with `PMM_QA_GIT_BRANCH` pointed here, is what provides it.

## Not fixed here

- **The underlying `pmm-admin list` rc=1 / `Service with ID … not found`.** Intermittent, fires right after the mongodb and postgresql socket services are removed. Could be a remove/list ordering race in the test or a `pmm-admin` bug; **not reproduced, so not classified.** This PR makes the next occurrence say so out loud, which is the precondition for diagnosing it.
- **The apt side of the publish race.** `apt-get update` already retries 5 × 10 s and still lost on Ubuntu 24.04 twice, because a retry re-uses the cached `Release` in `/var/lib/apt/lists`. Clearing that between attempts is the standard remedy, but it cannot be verified against a live republish, and tuning a retry blind is how a real signal gets buried.
- **`fail-fast: true` on the `setup` matrix** (`nightly-e2e-tests-matrix.yml:80`), which turns one slow shard into 13 cancelled shards. Declined for a good reason: `expected_setup_jobs: 14` would need reworking to tell "failed" from "not ready yet".


travagliad:
Both halves of that are right, and both were mine. Body corrected.

**The undescribed change.** `pmm3-client_integration_upgrade_custom_path.yml` does step `old_version` back a release when `pmm_version` is already the newest, and the body said "exactly these two fixes" while the diff carried three behavioural changes. It is now §3 with the four-case table, because it is the one change that only matters in combination with another PR: without Percona-Lab/jenkins-pipelines#4428 handing the arm64 tarball lanes the newest GA, the condition is unreachable and the change is a no-op. Omitting it made the diff look smaller than it is.

**No CI here covers `package_tests/**`.** Verified rather than taken on trust: all five package workflows — `package-test-fb`, `package-test-matrix`, `package-test-matrix-full`, `package-test-single`, `runner-package-test` — are `workflow_dispatch:` only, and nothing triggers on a push or pull request touching those paths. The 37 checks that did run here are the FB E2E suites, which exercise none of the six changed files. So a green PR here would say nothing about the seven-playbook blast radius in §1, and it would have been easy to read it as if it did.

The body now says that where the verification section starts, rather than leaving the local runs looking like a supplement to CI when they are in fact the only evidence. What is still missing, and named as such: which previously-green suites now fail on a metric that was always absent. That needs a `workflow_dispatch` of `package-test-matrix` against this branch, or a `pmm3-nightly-orchestrator` run with `PMM_QA_GIT_BRANCH` pointed here — the latter is planned as #4 regardless.

Worth adding for anyone reading this later: the two review catches on this round were both real defects in code I had already convinced myself was correct. The `shell` task returning only `makecache`'s status is one I looked straight at earlier today while checking whether `until` read the right `rc`, concluded "correct", and missed that a failing `clean all` vanishes — right question, wrong scope.

---

kiranvuyurru (approved):

PR #1374, opened 2026-09-08
Give two waits a budget that matches their source

travagliad:
Two lanes in `pmm3-nightly-orchestrator` #4 failed on a single test each while the rest of their suite passed. Both are waits sized for a fast source and applied to a slow one.

Raising a timeout is usually how a real failure gets buried, so each one is argued below rather than asserted.

## 1. `PMM-T743` — the shortest wait in the repo on the slowest source

`ui / @instances`, 46 of 47 tests green. The one failure:

```
Metrics "azure_memory_percent_average" is empty:
tried to check for 10 second(s) with 1 second(s) with interval
```

| | |
|---|---|
| Is the metric name wrong? | **No** — PMM's own dashboards reference `azure_memory_percent_average` 20 times |
| Is the Azure instance dead? | **No** — `PMM-T746`/`PMM-T748` assert `count > 0` on QAN for that same Azure service, and passed in the same run |
| Wait used here | `waitForMetric(current.metricName, null, 10)` |
| `waitForMetric` default | 30 s |
| What the other slow paths use | **90 s** — `pg_up`, `node_memory_MemTotal_bytes` |
| PMM's low-resolution scrape | **60 s** (`managed/models/settings.go`, `MetricsResolutions.LR`) |

So the test allows 10 seconds for a metric whose scrape interval alone is 60, before Azure Monitor's own one-minute granularity and ingestion lag. It cannot span even one guaranteed sample. The other entry in the same `DataTable`, `mysql_global_status_max_used_connections`, comes from `mysqld_exporter` scraping the instance directly and appears in under a second — which is why 46 of 47 pass and this one does not.

Corroborating, in the same file:

```js
// Removing to avoid failures due to missing metric, seems to be some changes on Azure deployments.
// metrics.add(['mysql_global_variables_azure_ia_enabled']);
```

Same symptom, worked around by deleting the assertion rather than diagnosing it.

**Why this isn't masking:** the metric name is verified correct and the instance is verified alive, so there is no known defect being hidden — only a budget below the floor set by PMM's own scrape interval. 90 s matches what this repo already gives its other slow paths. And it stays falsifiable: if it still fails at 90 s, the Azure Monitor integration is genuinely broken and the next step is credentials and the exporter, not the timeout.

## 2. The backup `Before` hook routes an operation to a replica set it just restarted

`upgrade / 3.8.1 OTHERS`, 31 of 32 tests green. The one failure is the `"before each" hook` for the MongoDB restore scenarios: `Server selection timed out after 30000 ms`.

```js
await I.verifyCommand('docker exec rs101 systemctl start mongod');

const c = await I.mongoGetCollection('test', 'test');

await c.deleteMany({ number: 2 });
```

`mongoGetCollection` does not connect — it returns `this.client.db(...).collection(...)`, and the driver connects lazily. So `deleteMany` is the first operation after the restart, and it is what performs server selection. `systemctl start` returns as soon as systemd accepts the unit; the replica set still has to elect a primary, and the driver has one 30-second budget to find one.

The delete is now attempted up to three times, sleeping `I.wait(10)` between attempts and throwing from the third — so a genuinely unreachable replica set still fails the hook, and fails it no later than before, just not on the first cold election.

Worth flagging beyond this PR: **all four `systemctl start mongod` call sites in that file have the same shape**, and the other three are inside scenario bodies. Only the `Before` hook has been observed failing, so only it is changed here; the pattern is worth a sweep.

## Verification

- `node --check` clean on both files.
- **eslint against the repo's own `.eslintrc` (`airbnb`), before and after**: identical findings — 25 `no-undef` (pre-existing; codeceptjs globals are not declared for an isolated invocation), 1 `no-multiple-empty-lines`, 1 `max-len`.
- The `waitForMetric` comparison is from the source: default `timeOutInSeconds = 30` in `grafanaAPI.js`, and every other call site in the repo enumerated.

### What CI here does and does not cover

An earlier version of this section claimed neither change is exercised by CI on this PR. **That was wrong for the MongoDB half**, and thanks to review for catching it:

- **`@bm-mongo` does run here.** `e2e-tests-matrix.yml:211` calls `fb-e2e-suite.yml`, whose `@bm-mongo` lane (`fb-e2e-suite.yml:76`) exercises the scenarios sitting behind the `Before` hook this PR changes. So change 2 can be proven green on this PR rather than waiting for the next nightly.
- **`@instances` does not.** Change 1 still needs a real PMM server with a real Azure instance, so the next `pmm3-nightly-orchestrator` remains its only signal.
- **No CI gate lints either `.js` file in this diff.** `Lint` is green, but `.claude/hooks/lint-changed.sh` invokes eslint only for TypeScript workspaces (line 63, gated on `ts_files`); the `js|ts` list it builds at line 52 feeds only a `skip-until:` date check at line 161. The airbnb run above is mine, not a check.

**Neither change is proof of a fix.** #1 assumes the metric arrives at all; #2 assumes the election finishes inside ~20 s of added budget. If either assumption is wrong the lane fails again, with the same message, and that is the useful next data point.

## Not fixed here

The two remaining `pmm3-nightly-orchestrator` #4 failures are a different class and get no speculative patch: `pmm-admin` returning empty inventory for a service that exists — `Service with ID … not found` on Debian 13, and `42008 != ` where `ps` sees the exporter but `pmm-admin` reports no port on Ubuntu 24.04. One OS in eight each, immediately after a service operation. They need reproduction with `pmm-agent` in debug, not a timeout.

### DIFF
--- a/codeceptjs-e2e/tests/backup/inventory_test.js
+++ b/codeceptjs-e2e/tests/backup/inventory_test.js
@@ -97,7 +97,28 @@ Before(async ({
 
   const c = await I.mongoGetCollection('test', 'test');
 
-  await c.deleteMany({ number: 2 });
+  // systemctl returns as soon as systemd accepts the unit, but the replica set
+  // still has to elect a primary before an operation can be routed to one, and
+  // this is the first operation after the restart. One server-selection budget
+  // is not always enough for a cold election, so give it three.
+  let selectionError;
+
+  for (let attempt = 0; attempt < 3; attempt += 1) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      await c.deleteMany({ number: 2 });
+      selectionError = undefined;
+      break;
+    } catch (error) {
+      selectionError = error;
+      // eslint-disable-next-line no-await-in-loop
+      await new Promise((resolve) => { setTimeout(resolve, 10000); });
+    }
+  }
+
+  if (selectionError) {
+    throw selectionError;
+  }
 
   await I.Authorize();
   await settingsAPI.changeSettings({ backup: true });
--- a/codeceptjs-e2e/tests/backup/inventory_test.js
+++ b/codeceptjs-e2e/tests/backup/inventory_test.js
@@ -97,10 +97,6 @@ Before(async ({
 
   const c = await I.mongoGetCollection('test', 'test');
 
-  // systemctl returns as soon as systemd accepts the unit, but the replica set
-  // still has to elect a primary before an operation can be routed to one, and
-  // this is the first operation after the restart. One server-selection budget
-  // is not always enough for a cold election, so give it three.
   let selectionError;
 
   for (let attempt = 0; attempt < 3; attempt += 1) {
--- a/codeceptjs-e2e/tests/backup/inventory_test.js
+++ b/codeceptjs-e2e/tests/backup/inventory_test.js
@@ -97,23 +97,19 @@ Before(async ({
 
   const c = await I.mongoGetCollection('test', 'test');
 
-  let selectionError;
-
   for (let attempt = 0; attempt < 3; attempt += 1) {
     try {
       // eslint-disable-next-line no-await-in-loop
       await c.deleteMany({ number: 2 });
-      selectionError = undefined;
       break;
     } catch (error) {
-      selectionError = error;
-      // eslint-disable-next-line no-await-in-loop
-      await new Promise((resolve) => { setTimeout(resolve, 10000); });
-    }
-  }
+      if (attempt === 2) {
+        throw error;
+      }
 
-  if (selectionError) {
-    throw selectionError;
+      // eslint-disable-next-line no-await-in-loop
+      await I.wait(10);
+    }
   }
 
   await I.Authorize();
--- a/codeceptjs-e2e/tests/verifyAzureMySQLPostgreSQLRemoteInstance_test.js
+++ b/codeceptjs-e2e/tests/verifyAzureMySQLPostgreSQLRemoteInstance_test.js
@@ -98,6 +98,6 @@ Data(filters).Scenario('PMM-T746, PMM-T748 - Verify adding monitoring for Azure 
 Data(metrics).Scenario(
   'PMM-T743 - Check metrics from exporters are hitting PMM Server @instances',
   async ({ grafanaAPI, current }) => {
-    await grafanaAPI.waitForMetric(current.metricName, null, 10);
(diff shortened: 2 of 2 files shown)


codeceptjs-e2e/tests/backup/inventory_test.js
```diff
@@ -97,7 +97,28 @@ Before(async ({
 
   const c = await I.mongoGetCollection('test', 'test');
 
-  await c.deleteMany({ number: 2 });
+  // systemctl returns as soon as systemd accepts the unit, but the replica set
+  // still has to elect a primary before an operation can be routed to one, and
+  // this is the first operation after the restart. One server-selection budget
+  // is not always enough for a cold election, so give it three.
```
travagliad:
```suggestion
```

travagliad:
Both findings verified and acted on. The second one is a straight correction to my own body — thanks.

**`@bm-mongo` coverage: you're right, I was wrong.** `e2e-tests-matrix.yml:211` → `fb-e2e-suite.yml:76` carries `tags_for_tests: '@bm-mongo'`, so the scenarios behind the changed `Before` hook do run on this PR. What makes it worse is that I had the evidence in hand and misread it: the check-run list on this PR literally contains `FB E2E tests / Backup Management MongoDB UI tests / e2e tests: @bm-mongo`, and I wrote the opposite anyway. Body corrected: change 2 is provable here, change 1 still needs the nightly.

**The `.js` lint gap: also right.** `.claude/hooks/lint-changed.sh` invokes eslint only at line 63, gated on `ts_files`; the `js|ts` list built at line 52 feeds only the `skip-until:` date check at line 161. So `Lint` being green says nothing about either file here, and the airbnb run in my body is mine rather than a gate. Stated as such now.

**The loop: applied in `6f0186ae`.** The sentinel did cost a real 10 seconds — on the third failure the `catch` still slept before the loop exited and the hook rethrew, so the change makes the hook fail *no later* than it did before while retrying twice. `I.wait(10)` is the house idiom; I counted **196** call sites across `codeceptjs-e2e/tests`, not 36, which only strengthens it. `I` was already destructured in the hook.

eslint against the repo's `.eslintrc` is identical to `origin/main` before and after: 25 `no-undef`, 1 `no-multiple-empty-lines`, 1 `max-len`. `node --check` clean.

One consequence worth naming: this push cancels the `@bm-mongo` run that was in flight on `6c69d1c1` via `cancel-in-progress`. That is the right trade — the run that matters is the one on the final head, and `6f0186ae` is it.

---

yurkovychv (approved):

PR #1408, opened 2026-09-10
PMM-T2287 Verify haproxy pods start without all pmm-ha pods running

shrutipradhan-percona:
Automates Zephyr **PMM-T2287**, covering **PMM-15394** — *[HA] HAProxy will not start unless every PMM replica is healthy*.

## The defect

The `pmm-ha` chart gave every HAProxy pod an init container (`wait-for-pmm-ready`) that polled *all* PMM replicas and `exit 1`-ed after 60 attempts. Because the StatefulSet is `OrderedReady`, the later replicas do not exist yet — no pod, no DNS record — so HAProxy sat at `Init:0/1` and CrashLoopBackOff'd instead of starting and returning 503 until a backend came up. The fix removes the init container; `haproxy.cfg` already handles the startup race with `init-addr last,libc,none` plus a `/v1/server/leaderHealthCheck` health check.

The contract restored by the fix, and what this test asserts: **HAProxy becomes Ready without every PMM replica being Ready.**

## The test

`e2e_tests/tests/ha/haproxyIndependentStartup.test.ts`, tagged `@pmm-ha`. Two independent oracles, so neither props up the other:

1. **Structural** — no HAProxy pod declares an init container. The fix *is* a removal, so its absence is worth asserting directly rather than inferring from timing. Guarded by `not.toHaveLength(0)` first, since a `for` loop over an empty pod list passes vacuously.
2. **Behavioural** — delete all PMM Server and HAProxy pods in one `kubectl delete`, then assert (a) the HAProxy pods reconverge while fewer than 3 PMM pods are Ready, and (b) the **last** HAProxy pod's Ready transition precedes the **first** PMM pod's, read from `.status.conditions[?(@.type=="Ready")].lastTransitionTime`.

Assertion (b) is computed from timestamps the API server already recorded, so unlike a live sample it cannot race. In (a), HAProxy is sampled *before* PMM inside the poll callback on purpose: the ~1s gap between the two reads can only let more PMM pods become Ready, so it can produce a false failure but never a false pass.

**Replacement pods are identified by UID, not name.** A terminating pod still reports `Ready=True` for its whole grace period, and a StatefulSet recreates `pmm-ha-0` under the same name — so counting by label alone would let the three pods on their way out satisfy the assertion.

`expect.poll` rather than `toPass`: nothing in the callback throws, since `kubectl get pods` keeps working throughout (unlike the leader-election helpers here that must tolerate HAProxy 5xx-ing mid-failover).

## Helper changes

| File | Change |
| --- | --- |
| `interfaces/kubernetes.ts` | `KubernetesPod` gains `initContainers`, `uid`, optional `readySince` |
| `helpers/k8s.helper.ts` | Populates those from the pod JSON; adds `deletePods(names)` so several pods go down in one call — `deletePod` delegates to it |
| `helpers/haCluster.helper.ts` | Exports `haproxyPodSelector` beside the existing `pmmServerPodSelector` |

`readySince` is `number | undefined` and set only when the Ready condition is true — a not-Ready pod's Ready condition still carries a `lastTransitionTime`, the moment it went *false*, and parsing that unconditionally would hand the ordering assertion a timestamp meaning the opposite of what it reads.

## Verification

Run on a throwaway Linode LKE cluster (4 × `g6-standard-4`, 3 PMM replicas, 3 HAProxy replicas, PMM Server 3.9.1). Rather than mutate an expected value to synthesise a failure, the test was run against the chart that actually carries the regression — the published chart is still buggy, so the fix went in as a chart upgrade mid-run:

| # | Chart | Result | Stopped on |
| --- | --- | --- | --- |
| 1 | `pmm-ha-1.6.1` | ✘ 10s | `initContainers` — expected `[]`, got `["wait-for-pmm-ready"]` |
| 2 | `pmm-ha-1.6.1` | ✘ 1m57s | `pmmReady` — expected `< 3`, got `3` |
| 3 | → `1.6.2` | upgraded | all HAProxy pods now report `initContainers=[]` |
| 4 | `pmm-ha-1.6.2` | ✓ 1m51s | — |

Run 2 had only the one structural expected value flipped to the buggy value, so the run reached the behavioural steps — which then caught the defect on their own evidence. Runs 1 and 4 are the same file byte for byte; only the chart changed.

Green-run readiness timeline after one simultaneous delete of all six pods — every HAProxy pod was serving **43s before the first** PMM pod was Ready:

```
pmm-ha-haproxy-…-tck58   11:04:00   +0s    initContainers=[]
pmm-ha-haproxy-…-xvv9v   11:04:00   +0s    initContainers=[]
pmm-ha-haproxy-…-9pmm2   11:04:01   +1s    initContainers=[]
pmm-ha-0                 11:04:44   +44s
pmm-ha-1                 11:05:11   +71s
pmm-ha-2                 11:05:37   +97s
```

`npm run lint` (eslint + `tsc --noEmit`) passes. Zephyr PMM-T2287 moved `Needs Automation` → `Automated`.

## Notes for reviewers

- **Nothing reports this to Zephyr yet.** The Playwright suite has no Zephyr reporter — only the legacy CodeceptJS suite posts executions. The key in the title is the traceability link and the repo convention, but this test will not appear in a Zephyr cycle until that reporter exists.
- **Both the bug and the fix live in `percona/percona-helm-charts`, not the server image** — `appVersion` was 3.9.1 on both sides. Running this case against the released chart tests the wrong thing.
- **Three worker nodes are not enough** for the `@pmm-ha` suite once chart 1.6.2 adds HAProxy resource requests: `pmm-ha-0` (4 GiB) then has nowhere to schedule.

Two chart issues were hit while verifying and are **not** addressed here, since they are outside this repo:

- The 1.6.1 → 1.6.2 HAProxy rollout can deadlock — the released chart's *required* `podAntiAffinity` on `kubernetes.io/hostname` pins one pod per node, so with the Deployment's default `maxUnavailable: 0` no replacement can schedule, while Helm still reports `STATUS: deployed`. Deleting the superseded ReplicaSet clears it. Same class as PMM-15393, which chart 1.6.2's own notes cite.
- A fresh HA install 401s on admin login: the chart's `pmm-token-init` job crash-loops and its retries keep Grafana's 5-minute brute-force lockout alive, so the documented password reset appears to do nothing — Grafana reports the lockout under the same `password-auth.failed` id as a wrong password. Suspending the job lets the window expire, after which the job completes on its own.

https://claude.ai/code/session_018bR4sxcXg7B9wj3j2tZ9hp

---
_Generated by [Claude Code](https://claude.ai/code/session_018bR4sxcXg7B9wj3j2tZ9hp)_


yurkovychv:
@shrutipradhan-percona  could you pls run ha-testing pipeline so we have a proof that the tests are green?

shrutipradhan-percona:
@copilot resolve the merge conflicts in this pull request

PR #1417, opened 2026-09-12
migrate(dashboards): verifyHomeDashboards CodeceptJS to Playwright

kiranvuyurru:
Migrates `verifyHomeDashboards` from CodeceptJS to native Playwright, retires the source, and gives the migrated scenarios a Playwright consumer in the nightly matrix.

- **Source:** `codeceptjs-e2e/tests/dashboards/verifyHomeDashboards_test.js`
- **Target:** `e2e_tests/tests/dashboards/homeDashboards.test.ts` (new file)

## Scenarios and tags

Every source tag is preserved verbatim, including the double space the source has between `@nightly` and `@dashboards`.

| Scenario | Tags | Shape |
| --- | --- | --- |
| PMM-T1565 - Verify ability to access OS dashboards with correct filter setup from Home Dashboard | `@nightly  @dashboards` | source `Data(panels).Scenario(...)` over a 7-row `DataTable` → `for` loop over a typed `PanelRow[]`, 7 generated tests |
| PMM-T2007 - Verify Monitored DB Services panel on home dashboard | `@nightly  @dashboards @gssapi-nightly` | single test |

**Titles are byte-identical to the CodeceptJS originals**, proven by an index-aligned diff of both runners' listings: **8/8 match, 0 differ**, equal byte lengths per index.

That byte-identity is why each row carries a hand-written `titleSuffix` literal rather than `JSON.stringify(panel)`. The initial review gate rejected the stringify form: `perfectionist/sort-objects` alphabetises the migrated object literal while `new DataTable([...])` keys its JSON in **column** order, and the migrated row carries the target POM key (`diskDetails`) where the source carries `osDiskDetails`. The literals hold the source row's own JSON — column order `panelName, dashboardType, dashboardName, dashboard`, with the `os*` keys. Please do not "simplify" them back into a stringify; it silently retitles all 7 generated tests.

### Retry scope

The source attaches `.retry(1)` to the PMM-T1565 `Data` scenario only; PMM-T2007 carries no retry. Reproduced with an anonymous `pmmTest.describe` wrapping just the PMM-T1565 loop plus `pmmTest.describe.configure({ retries: 1 })` inside it; PMM-T2007 sits outside and inherits the config default. The anonymous describe was verified **not** to prefix titles (JSON reporter suite title is `""`, and exact-title greps select exactly 1 test each).

## Bucket and environment

The tracker proposed bucket **B1/none**, which cannot run either scenario. Corrected to **B9**:

```
--database mysql --database psmdb --database pgsql     # setup_client: false
```

- PMM-T1565 needs **at least two monitored nodes** — it selects `nodeNames[0]` and `nodeNames[1]`.
- PMM-T2007 waits on the literal text `MySQL` in the Monitored DB Services panel.
- `setup_client: false` — no host `pmm-admin`/`pmm-agent` invocation anywhere in the source or its page objects.

The tracker's `Tags` cell was also wrong: it read `@dashboards`, omitting `@nightly` (the tag the nightly matrix actually selects on) and `@gssapi-nightly`.

## Execution evidence

Whole file, `--workers=1`, declaration order, against remote PMM **3.10.0-v3-8b6be9b8d** (6 nodes; mysql 1 / mongodb 3 / postgresql 2 / proxysql 0):

**8 passed (4.5m) — every test on its first attempt, zero retries.**

| Test | Duration |
| --- | --- |
| PMM-T1565 Disk Space Total | 33.3s |
| PMM-T1565 Disk Reads | 23.6s |
| PMM-T1565 Disk Writes | 22.8s |
| PMM-T1565 Total RAM | 27.3s |
| PMM-T1565 Virtual Memory Total | 49.7s |
| PMM-T1565 Monitored Nodes | 49.5s |
| PMM-T1565 Total Virtual CPUs | 50.1s |
| PMM-T2007 | 5.4s |

New target file, so this complete-file run is the required proof; no target-file regression run applies.

Re-run once more from the publish branch itself, immediately before pushing, to confirm the exact tree in this PR: **8 passed (5.0m)**, again every test on its first attempt — Disk Space Total 36.4s, Disk Reads 25.4s, Disk Writes 24.1s, Total RAM 30.7s, Virtual Memory Total 54.0s, Monitored Nodes 59.5s, Total Virtual CPUs 56.2s, PMM-T2007 6.0s.

### The environment story — worth reading before judging any red run

An earlier run on a **local Windows Docker host** failed 3 of 8 — all three `osDiskDetails` rows, all on the no-data budget and nothing else. **The root cause was block-device topology, not the test.** That host exposes 6 block devices with **two permanently idle** (`sdb`, `sdf` — zero disk ops at every window from 5m to 12h), while `osDiskDetails.noDataElements = 3` budgets one. Disk Details repeats its panels per device, so 2 idle devices × 3 activity-derived panels = 6 no-data panels against a threshold of 3.

Confirmed from both ends:

- The **unmigrated CodeceptJS source fails identically on that host** — `Expected 3 Elements without data but found 4`, same sentence, same assertion.
- Nightly CI run <https://github.com/percona/pmm-qa/actions/runs/34629383719> shows all 7 PMM-T1565 source rows **passing the same day**.

The verification host has 2 devices per node, both with data, **zero idle** → 0 no-data panels against a budget of 3.

## Workflow coverage

One line, in `.github/workflows/nightly-e2e-tests-matrix.yml`:

```diff
-          - tags_for_tests: '@annotations|@menu'
+          - tags_for_tests: '@annotations|@menu|@dashboards'
```

Without it the two migrated scenarios are selected by **no** Playwright job, and the coverage the retired source provided under `test execution / @nightly` disappears silently on retirement.

`expected_test_jobs: 3` (line 115) is **unchanged and still correct** — appending a tag to an existing matrix entry spawns no job. YAML-parsed count: 2 `test_execution` entries + 1 `test_execution_playwright` entry = 3 before and after. That file's own comment (lines 141-145) prescribes exactly this append for exactly this reason.

Selection, measured with `npx playwright test --list --grep`:

| Expression | Tests |
| --- | --- |
| `'@annotations\|@menu'` (before) | 7 in 2 files |
| `'@dashboards'` | 25 in 4 files |
| `'@annotations\|@menu\|@dashboards'` (after) | **26 in 5 files** |

**+19 = 8 this PR's + 10 valkey PMM-T2087 + 1 PMM-T2262.**

### ⚠️ Valkey consequence — please accept or revert this knowingly

The 10 `PMM-T2087` tests in `e2e_tests/tests/dashboards/valkey/valkeyDashboards.test.ts` are tagged `@nightly @dashboards @pmm-valkey-integration`. **Their only consumer, the `valkey_integration` job in `.github/workflows/e2e-tests-matrix.yml` (lines 161-171), is entirely commented out** — and `git log -L 161,171` shows it was added *already commented* in `96a45a54`. So `@pmm-valkey-integration` has never had an active consumer, and **this edit switches those 10 tests on in CI for the first time**, on a job whose failures fire the Investigator routine.

They do have their service there: the nightly `ps-gr-pxc-valkey` setup shard provisions `--database valkey`.

This was an explicit decision in preference to introducing a narrower tag. Reverting is one line if you disagree.

The 11th, `PMM-T2262` MongoDB Unused Indexes (`@nightly @dashboards @pmm-psmdb-integration`), also runs in `e2e-tests-matrix.yml`'s psmdb job; the `mysql-psmdb-pgsql` shard provisions `--database psmdb,SETUP_TYPE=pss`.

### Coverage-edit execution evidence

The edited job's **own full expression**, at the job's own worker count:

```
npx playwright test --grep '@annotations|@menu|@dashboards' --workers=1
```

**12 passed / 14 failed (7.0m). All 8 migrated tests passed; none of the 14 failures comes from this file.**

The control that makes those failures interpretable: **3 of the 14 were already selected by the pre-edit expression** — the baseline `'@annotations|@menu'` scores only 4/7 on this host. So the failures measure the ad-hoc verification environment, not this change:

- 10 × `Service type valkey is not present` — no valkey service on this host (the nightly shard provisions one).
- 1 × `MongoTopologyClosedError: Topology is closed` in `dropDatabase` — a *direct* MongoDB connection from a Windows host to the remote replica set, not a PMM assertion.
- 3 × `Service ps_ of type mysql is not present` — pre-existing, in the job before this edit.

### Reverse direction

Bounded first: `git diff --name-status -M origin/main HEAD -- e2e_tests/tests/` shows a single addition and no modified test file, so no existing expression's selection can change except by newly matching the added file. Then all 16 existing Playwright job expressions were listed — **each selects 0 tests from `homeDashboards.test.ts`**: `@annotations|@menu` (7), `@new-navigation|@menu` (24), `@pmm-psmdb-integration` (1), `@inventory|@image-renderer` (5), `@settings` (14), `@LBAC` (1), `@pmm-ps-pxc-haproxy-integration` (9), `@pmm-valkey-integration` (10), `@fb-settings` (3), `@service-account` (3), `@user-password` (1), `@alerting` (5), `@rta` (21), `@docker-configuration` (6), `@fb-instances` (2), `@pmm-ps-integration` (10).

## Accepted deviation — assertion strictness

The source's `verifyThereAreNoGraphsWithoutData` counts no-data panels with `grabNumberOfVisibleElements` (**visible only**); the migration counts with `noDataPanelName.allTextContents()` (**all attached**). Measured 6 vs 4 on the old host, 0 vs 0 on the verification host.

The divergence is **one-directional** — the migration can only ever count the same or more, never fewer.

**Residual risk:** on a CI host where the source counts exactly at its budget, an off-screen no-data panel could fail the migration where the source passes.

This is a deliberate decision to ship as-is rather than change the assertion inside a migration PR.

## Retirement

`git mv` to `codeceptjs-e2e/tests/dashboards/verifyHomeDashboards_migrated.js` — **R100, zero content change**. `codeceptjs-e2e/pr.codecept.js:141` discovers `tests/**/*_test.js`, so a `_migrated.js` file is no longer collected.

No job is emptied:

- `@nightly` (selected by `test execution / @nightly`): **53 occurrences across 14 active `*_test.js` → 51 across 13**.
- `@dashboards` has no `tags_for_tests` consumer in any workflow file.
- `@gssapi-nightly` has no consumer either — preserved verbatim on the migrated test, but it is not a selection route and never was.

Page objects are kept, per the rows 5/6/7/8 precedent. `codeceptjs-e2e/tests/pages/dashboards/homeDashboard.js` still has 2 callers (`tests/configuration/permissions_test.js`, `tests/verifyPMMDemoDashboards_test.js`), and `dashboardPage.js` is repo-wide.

## Files queried from the existing graphs

Both graphs were refreshed on the control branch before the migration and queried (not regenerated) to derive the dependency lists: Playwright **827 → 892 nodes / 1851 → 2035 edges**; CodeceptJS **2095 nodes / 3101 edges, net unchanged**. Neither graph is in this PR.

## CI

E2E tests Matrix run for this branch: https://github.com/percona/pmm-qa/actions/runs/34661128973

## Validation

| Check | Result |
| --- | --- |
| `eslint .` | clean |
| `tsc --noEmit` | clean |
| `python support_scripts/generate_readme.py --check` | up to date (`@gssapi-nightly` is new to the Playwright suite, so `e2e_tests/README.md` is regenerated here) |
| migration conventions script | passed |
| MCP locator verification | **clean, zero locator fixes needed** |

MCP verification detail: `monitoredServicesCounts`' promoted testid chain returns the identical 5 spans in the identical order as the source's CSS form; all 7 clicked panels have exactly 1 header link and 1 Data link, so `getByRole('link')` is not a strict-mode hazard; `dashboardTitle` resolves to 1 breadcrumb span on both OS dashboards.

## Review history

| Gate | Verdict |
| --- | --- |
| Initial review, attempt 1 | `REVIEW_FAILED` — B1, data-row title byte-identity |
| Initial review, attempt 2 | `READY_TO_RUN` |
| Final review, attempt 1 | `FINAL_REVIEW_FAILED` — B2, commit-body prose |
| Final review, attempt 2 | `FINAL_REVIEW_FAILED` — B3, commit-body prose |
| Final review, attempt 3 | `FINAL_REVIEW_PASS` |

B2 and B3 were both defects in **commit body prose, not code** — this repository squash-merges and the squash body is the concatenation of branch commit messages, so a false clause in a commit body reaches `main` verbatim. Both were fixed by message-only history rewrites; all three tree SHAs were proven byte-identical across every rewrite.

## Open advisory for a reviewer

**A13:** `acceptableNoDataCount` is used by `diskDetails.ts`, `memoryDetails.ts` and `nodesOverview.ts` but is **not declared on `e2e_tests/interfaces/dashboard.ts`**. Left out of scope here deliberately — it is a code change to a shared interface and did not belong in a migration PR under review. Worth a follow-up.


kiranvuyurru:
GitHub Actions: https://github.com/percona/pmm-qa/actions/runs/34661128973

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,205 @@
...
+          )
+          .toBe(true);
+        await dashboard.verifyMetricsPresent(osDashboard.metrics);
+
+        const noDataPanels = await dashboard.elements.noDataPanelName.allTextContents();
+        const realFailures = noDataPanels.filter(
+          (title) => !title.toLowerCase().includes('24') && !title.toLowerCase().includes('hour'),
+        );
+        const unacceptableNoDataPanels =
+          noDataPanels.length > osDashboard.acceptableNoDataCount ? realFailures : [];
```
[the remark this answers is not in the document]
kiranvuyurru: Not changed - this is deliberate fidelity to the source, not an oversight. Both behaviours you describe are the original test's: in `codeceptjs-e2e/tests/pages/dashboardPage.js:1341-1355` the only assertion (`printFailedReportNames`) sits inside **both** guards, so the source fails only when `numberOfNAElements > acceptableNACount` AND `realFailures.length > 0`. This PR is a CodeceptJS-to-Playwright migration, and the migration rules forbid adding, removing or strengthening coverage - asserting `realFailures` independently would make the migrated test strictly stricter than the source and could fail nightly where the source passes. The observation is a fair criticism of the original assertion and worth a separate ticket against it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,205 @@
...
+              ),
+            {
+              message: `${panel.dashboardName} should keep the Node Name filter "${expectedNodeNames}" selected`,
+              timeout: Timeouts.TWENTY_SECONDS,
+            },
+          )
+          .toBe(true);
+        await dashboard.verifyMetricsPresent(osDashboard.metrics);
+
+        const noDataPanels = await dashboard.elements.noDataPanelName.allTextContents();
```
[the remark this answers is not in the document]
kiranvuyurru: Measured on a live PMM rather than reasoned about, and the premise does not hold: `allTextContents()` does not undercount here. After `loadAllPanels()` (which `verifyMetricsPresent` calls) every grid item has been scrolled through and stays mounted, so a single call sees them all.

| dashboard | gridItems | panelHeaders raw | scrolled | noData raw | noData scrolled |
| --- | --- | --- | --- | --- | --- |
| Nodes Overview | 91 | **81** | **79** | 0 | 0 |
| Memory Details | 44 | 38 | 38 | **5** | **5** |
| Disk Details | 30 | 31 | 31 | 0 | 0 |

Raw is equal or **higher** in every case, never lower. The gap on Nodes Overview runs the other way for the reason you flagged yourself: `collectTextsAcrossScroll` dedupes through a `Set`, and Nodes Overview has genuinely duplicate panel titles, so switching to it would drop 81 to 79 - shrinking the count toward `acceptableNoDataCount` and weakening the threshold, which is the failure mode you were trying to prevent. The initial review gate rejected that swap for the same reason.

Memory Details is the useful control: 5 no-data panels against a budget of 6, both methods agreeing exactly, so the budget is live and near its limit rather than unreachable. Keeping the raw call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,205 @@
...
+        const osDashboard = dashboard.os[panel.dashboard];
+        const panelDataLink = dashboard.builders
+          .panelByExactName(panel.panelName)
+          .getByTestId('data-testid Data link');
+
+        await page.goto(urlHelper.buildUrlWithParameters(dashboard.home.url, { from: 'now-12h', to: 'now' }));
+        await dashboard.waitForDashboardToLoad();
+
+        const nodeNames = (await dashboard.getVariableValues('Node Name')).filter((name) => name !== 'All');
+        const currentPanelValue = await panelDataLink.innerText();
```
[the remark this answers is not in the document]
kiranvuyurru: Applied in 59bb3141, and I verified the mechanism you describe before taking it: `selectVariableValue` with an absent value falls through to `options.filter({ hasText: /^(?!All$).+/i })` (`pages/base.page.ts:87-93`), so on a one-node environment it re-clicks `nodeNames[0]` and deselects it rather than failing, and the run dies 30s later on the unrelated `toHaveCount(0)` wait.

Taken as a precondition guard rather than added coverage: on any environment the scenario is meant to run on it is a no-op and cannot change a passing verdict - it only makes a wrong environment name itself.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  },
+  {
+    dashboard: 'nodesOverview',
+    dashboardName: 'Nodes Overview',
+    dashboardType: 'multipleNodes',
+    panelName: 'Total Virtual CPUs',
+    titleSuffix:
+      '{"panelName":"Total Virtual CPUs","dashboardType":"multipleNodes","dashboardName":"Nodes Overview","dashboard":"osNodesOverview"}',
+  },
+];
```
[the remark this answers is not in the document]
kiranvuyurru: Not taking this one, though the maintenance point is fair. The `titleSuffix` literals are not incidental - the initial review gate failed this PR on exactly the construction you are proposing. `JSON.stringify(panel)` cannot reproduce the CodeceptJS suffix: `perfectionist/sort-objects` alphabetises the migrated literal while CodeceptJS keys the JSON in `new DataTable([...])` column order, and the migrated row carries the target POM key (`diskDetails`) where the source carried `osDiskDetails`. The first attempt shipped the stringify form and 7 of 8 titles diverged; the literals brought it to 8/8, proven by an index-aligned diff of both runners' listings.

You are right that the byte-identity property is not consumed on the Playwright path - its purpose was to prove at migration time that no scenario or data row was silently dropped or renamed, and it has now served that purpose. The migration rules do allow a clearer title on an explicit reviewer request once both gates have passed. But that is a maintainer's call to make deliberately, in its own commit, rather than something to fold into a review round - so I am leaving it and flagging it for @kiranvuyurru instead of reversing a gate decision unilaterally.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,205 @@
...
+pmmTest.beforeEach(async ({ grafanaHelper }) => {
+  await grafanaHelper.authorize();
+});
+
+pmmTest.describe(() => {
+  pmmTest.describe.configure({ retries: 1 });
+
+  for (const panel of panels) {
+    pmmTest(
+      `PMM-T1565 - Verify ability to access OS dashboards with correct filter setup from Home Dashboard @nightly  @dashboards | ${panel.titleSuffix}`,
```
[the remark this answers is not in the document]
kiranvuyurru: Declined together with the companion thread on the `panels` table - see the reasoning there. Dropping `titleSuffix` and deriving the suffix with `JSON.stringify(panel)` is precisely what the initial review gate failed this PR for: the lint rule alphabetises the literal and the row carries `diskDetails` where the source carried `osDiskDetails`, so 7 of the 8 titles diverge.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/nightly-e2e-tests-matrix.yml
```diff
@@ -153,7 +153,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - tags_for_tests: '@annotations|@menu'
+          - tags_for_tests: '@annotations|@menu|@dashboards'
```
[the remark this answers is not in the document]
kiranvuyurru: Fair catch on the evidence, and you are right about the linked run: `E2E tests Matrix` does not exercise `nightly-e2e-tests-matrix.yml`, so it is not evidence for this edit and should not have been offered as though it were.

The widening itself is a deliberate maintainer decision by @kiranvuyurru, taken with the valkey consequence on the table - the alternative of a narrower destination tag was considered and rejected. Your description of the state is accurate and worth restating for whoever reviews next: `valkey_integration` in `e2e-tests-matrix.yml` was added already commented out, so `@pmm-valkey-integration` has never had an active consumer, and this appends those 10 `PMM-T2087` tests to a nightly job whose failures fire the Investigator routine.

Leaving the tag as-is per that decision rather than reverting it in a review round. Flagging the split-or-prove-first option back to @kiranvuyurru, since it is the same call and the evidence gap you identified is real.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/operating-system/memoryDetails.ts
```diff
@@ -0,0 +1,50 @@
...
+  'Pages Mapped by Files',
+  'Kernel Stack Pages',
+  'Slab Pages',
+  'Allocations',
+  'Refill',
+  'Direct Scan',
+  'Kswapd Scan',
+  'Steal Direct',
+  'Steal Kswapd',
+];
```
travagliad: There is no real gain to have this, why not inline it as we have done with others dashboards?

e2e_tests/pages/dashboards/operating-system/diskDetails.ts
```diff
@@ -0,0 +1,20 @@
...
+const panelNames = [
+  'Mountpoint Usage',
+  'Disk Latency',
+  'Disk Operations',
+  'Disk Bandwidth',
+  'Disk Load',
+  'Disk IO Utilization',
+  'Avg Disks Operations Merge Ratio',
+  'Disk IO Size',
+];
```
travagliad: There is no real gain to have this, why not inline it as we have done with others dashboards?

e2e_tests/pages/dashboards/operating-system/nodesOverview.ts
```diff
@@ -1,7 +1,93 @@
...
+  'Top Retransmit rate',
+  'Top 5 Network Traffic',
+  'Network Traffic',
+  'Top 5 Local Network Errors',
+  'Errors',
+  'Top 5 TCP Retransmission',
+  'Retransmission',
+  'Top 5 Local Network Drop',
+  'Drop',
+];
```
travagliad: There is no real gain to have this, why not inline it as we have done with others dashboards?

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      '{"panelName":"Monitored Nodes","dashboardType":"multipleNodes","dashboardName":"Nodes Overview","dashboard":"osNodesOverview"}',
+  },
+  {
+    dashboard: 'nodesOverview',
+    dashboardName: 'Nodes Overview',
+    dashboardType: 'multipleNodes',
+    panelName: 'Total Virtual CPUs',
+    titleSuffix:
+      '{"panelName":"Total Virtual CPUs","dashboardType":"multipleNodes","dashboardName":"Nodes Overview","dashboard":"osNodesOverview"}',
+  },
```
travagliad: I know this is included for verbatim migrations, but this achieves 2:

1. Big test name that can't be rendered on artifacts
2. repeating of above informations

Can we just adapt the test to include "panelName" somewhere in there and let it be it?

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+type OsDashboardKey = 'diskDetails' | 'memoryDetails' | 'nodesOverview';
+
+interface PanelRow {
+  dashboard: OsDashboardKey;
+  dashboardName: string;
+  dashboardType: 'multipleNodes' | 'singleNode';
+  panelName: string;
+  titleSuffix: string;
+}
```
travagliad: Interface isn't needed, nowhere this is being used

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,208 @@
+import { expect } from '@playwright/test';
+import pmmTest from '@fixtures/pmmTest';
+import { GetService, ServiceType } from '@interfaces/inventory';
+import { Timeouts } from '@helpers/timeouts';
+
+type OsDashboardKey = 'diskDetails' | 'memoryDetails' | 'nodesOverview';
```
travagliad: This isn't needed

e2e_tests/tests/dashboards/homeDashboards.test.ts
travagliad: This can be very reduced to the point of only specifying panelName that's the only real unique value, then other params can eiither come from page or just don't exist at all, a lot of redundancies in the test code too

.github/workflows/nightly-e2e-tests-matrix.yml
travagliad: why not @nightly ?

travagliad (changes requested):

e2e_tests/pages/dashboards/operating-system/memoryDetails.ts
kiranvuyurru: Done in 8eb90420 - the `panelNames` const and the `.map()` are gone, and `metrics` is declared inline in the same shape `nodeSummary.ts` and the other dashboards in this folder use. 38 entries here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/operating-system/diskDetails.ts
kiranvuyurru: Done in 8eb90420 - the `panelNames` const and the `.map()` are gone, and `metrics` is declared inline in the same shape `nodeSummary.ts` and the other dashboards in this folder use. 8 entries here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/operating-system/nodesOverview.ts
kiranvuyurru: Done in 8eb90420 - the `panelNames` const and the `.map()` are gone, and `metrics` is declared inline in the same shape `nodeSummary.ts` and the other dashboards in this folder use. 81 entries here.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
kiranvuyurru: Done in 8eb90420. The title now ends in the one value that actually varies:

`PMM-T1565 - ... @nightly  @dashboards | Disk Space Total`

The `titleSuffix` field it used to interpolate is deleted outright. Worth recording why it existed: the migration gates require the generated titles to be byte-identical to the CodeceptJS originals, as evidence that no scenario or data row was silently dropped or renamed - the first attempt at this PR was in fact failed by that check. That evidence has served its purpose now that both gates have passed, so spending it on readability is the right trade. The parts that are actually parsed - every tag and the `PMM-T1565` token that `fixtures/pmmTest.ts` uses to drive its version gate - are unchanged.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
kiranvuyurru: Removed in 8eb90420, along with the `OsDashboardKey` alias on line 6. `as const` on the array gives the same literal typing without either declaration.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
kiranvuyurru: Removed in 8eb90420, together with the `PanelRow` interface below it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
kiranvuyurru: Done in 8eb90420 - the header drops from 73 lines to 14, and the file loses 89 lines net.

Rows went from five fields to three:

```ts
const panels = [
  { dashboard: 'diskDetails', dashboardName: 'Disk Details', panelName: 'Disk Space Total' },
  ...
] as const;
```

`dashboardType` is gone: it was fully determined by `dashboard` - every diskDetails and memoryDetails row was `singleNode` and every nodesOverview row `multipleNodes` - so it is derived at the call site as `panel.dashboard !== 'nodesOverview'`. `titleSuffix` and the interface/alias are gone too.

I kept `dashboardName` rather than deriving it. It is the argument to one of the source's own assertions (the post-navigation dashboard title check), so it has to exist somewhere; the only way to get it "from the page" would be to title-case the URL slug (`node-disk/disk-details` -> `Disk Details`), which happens to work for all three but is string-munging I would rather not rely on. Happy to move it onto the page objects instead if you prefer that home for it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/nightly-e2e-tests-matrix.yml
kiranvuyurru: Because `@nightly` is the wider of the two, not the narrower. Measured on this branch:

| grep | selects |
| --- | --- |
| `@nightly` | 27 tests in 6 files |
| `@dashboards` | 25 tests in 4 files |

The extra that `@nightly` picks up is `PMM-T2095 - Verify all agents for Mysql have status Running @nightly @pmm-ps-integration` - a test owned by the PS integration job, whose services these setup shards do not provision. That is the case this workflow's own comment above the job warns about: "Not `@nightly` -- here it also marks tests owned by other jobs whose services these shards never provision."

Neither tag is clean, to be fair: both also select the 10 `PMM-T2087` valkey tests, which is called out in the PR body.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

travagliad: So the appropriate fix would be to have @nightly on actual nightly tests, and clone @valkey-nightly @mongodb-nightly that we have for codeceptjs, so we can just say @nightly and run the exact same subset

e2e_tests/tests/dashboards/homeDashboards.test.ts
```diff
@@ -0,0 +1,149 @@
...
+import { Timeouts } from '@helpers/timeouts';
+
+const panels = [
+  { dashboard: 'diskDetails', dashboardName: 'Disk Details', panelName: 'Disk Space Total' },
+  { dashboard: 'diskDetails', dashboardName: 'Disk Details', panelName: 'Disk Reads' },
+  { dashboard: 'diskDetails', dashboardName: 'Disk Details', panelName: 'Disk Writes' },
+  { dashboard: 'memoryDetails', dashboardName: 'Memory Details', panelName: 'Total RAM' },
+  { dashboard: 'nodesOverview', dashboardName: 'Nodes Overview', panelName: 'Virtual Memory Total' },
+  { dashboard: 'nodesOverview', dashboardName: 'Nodes Overview', panelName: 'Monitored Nodes' },
+  { dashboard: 'nodesOverview', dashboardName: 'Nodes Overview', panelName: 'Total Virtual CPUs' },
```
travagliad: dashboard and dashboardName looks like repetitions just to format the text, can't we figure another way here to make this 1 property only?

.github/workflows/nightly-e2e-tests-matrix.yml
kiranvuyurru: Agreed on the direction - that is the right end state, and it is what makes `@nightly` mean something on the Playwright side rather than being a second name for "some dashboards plus some integration tests".

Here is the full inventory, so the scope is visible. `@nightly` currently selects 27 Playwright tests:

| tests | id | tags | owning job |
| --- | --- | --- | --- |
| 7 | PMM-T1565 | `@nightly  @dashboards` | this PR - genuinely nightly |
| 1 | PMM-T2007 | `@nightly  @dashboards @gssapi-nightly` | this PR - genuinely nightly |
| 6 | PMM-T878 | `@nightly  @dashboards @annotations` | already in this job via `@annotations` |
| 10 | PMM-T2087 | `@nightly @dashboards @pmm-valkey-integration` | valkey (its job is commented out) |
| 1 | PMM-T2262 | `@nightly @dashboards @pmm-psmdb-integration` | PSMDB integration |
| 1 | PMM-T2095 | `@nightly @pmm-ps-integration` | PS integration |
| 1 | PMM-T2030 | `@nightly @pmm-ps-integration` | PS integration |

So the retag set is the bottom four rows - **13 tests across four files, none of which this PR touches** (`valkey/valkeyDashboards.test.ts`, `mongo/mongodbUnusedIndexes.test.ts`, and the two PS integration specs). Once those carry `@valkey-nightly` / `@mongodb-nightly` / a PS equivalent instead of a bare `@nightly`, this job greps `@nightly` and gets exactly the 14 that belong to it - our 8 plus PMM-T878's 6 - which is the subset these setup shards actually provision for.

I have not made that change here, because retagging 13 tests in other people's specs from inside a dashboards migration is the kind of thing that is much easier to review and revert on its own. Happy to open it as a follow-up PR straight away if you want it, and then switch this line to `@nightly` either in that PR or in a one-line follow-up to this one.

For the interim, both options over-select and `@dashboards` is the narrower of the two: `@dashboards` = 25 tests, `@nightly` = 27 (the two extra being the PS integration pair above). Neither avoids the 10 valkey tests - only the retag does.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/homeDashboards.test.ts
kiranvuyurru: Done in a28bde8d. The display name moved onto the dashboard itself - it sits next to `url` on each OS page object, and the test reads `osDashboard.name` - so the rows now carry only what actually varies:

```ts
const panels = [
  { dashboard: 'diskDetails',   panelName: 'Disk Space Total' },
  { dashboard: 'memoryDetails', panelName: 'Total RAM' },
  { dashboard: 'nodesOverview', panelName: 'Monitored Nodes' },
  ...
] as const;
```

That also removes the drift risk you were pointing at - the key and the title can no longer disagree, because there is only one of them.

One deliberate omission: I did **not** add `name` to `DashboardInterface`. Required would break every page object implementing it; optional would advertise a convention nothing else follows yet. Happy to promote it repo-wide as its own change if you want that.

Verified on a locally provisioned PMM (qa-integration; 6 nodes, mysql 1 / mongodb 3 / postgresql 2): **5 passed / 3 failed**. The three are the Disk Details rows on the no-data budget, and they are environmental rather than a regression - Disk Details repeats its panels per block device and its ratio panels divide by completed operations, so an idle device produces no series at all. I measured the node those rows select as having 3 idle devices of 6 before running, which predicts 9 no-data panels against a budget of 3, and the run reported exactly 9. The same three rows pass on a host with no idle devices, and the unmigrated CodeceptJS source fails on them here too.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

.github/workflows/nightly-e2e-tests-matrix.yml
kiranvuyurru: Done in `212ec958` — the job now greps `@nightly`, with `PMM-T2087` retagged `@valkey-nightly` and `PMM-T2262`/`PMM-T2095`/`PMM-T2030` left to their own job tags; verified on a live PMM (15 tests in 3 files selected, this PR's 8 all green).

---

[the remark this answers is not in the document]
kiranvuyurru: Resolved in `212ec958` — those 10 tests are now `@valkey-nightly`, so this job no longer switches them on.

---

e2e_tests/tests/dashboards/homeDashboards.test.ts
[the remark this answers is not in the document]
kiranvuyurru: Superseded — `titleSuffix` and the `PanelRow`/`OsDashboardKey` declarations were removed in `8eb90420`, so the rows now carry only what varies.

---

e2e_tests/tests/dashboards/homeDashboards.test.ts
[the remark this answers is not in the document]
kiranvuyurru: Superseded — the title now ends in `| ${panel.panelName}` after `8eb90420`.

---

travagliad (approved):

PR #1419, opened 2026-09-13
Stop the dashboard metric walk failing on a Grafana render stall

travagliad:

- source: nightly CI run https://github.com/percona/pmm-qa/actions/runs/34727137335 (`test execution / @nightly`, `INSTALLATION_TYPE: docker`, `perconalab/pmm-server:3-dev-latest` = image `06f2a8fe42f2`)
- tests:
  - `codeceptjs-e2e/tests/dashboards/verifyPostgresqlDashboards_test.js` / `@nightly @dashboards` — PMM-T2048 — Verify PostgreSQL Instances Overview Extended metrics

64 of 65 tests passed in that job; PMM-T2048 was the only failure.

```
Error: element ([data-testid*="data-testid Panel header"][data-testid*="Total Blocks Operations"])
```

## The panel was there

`Total Blocks Operations` is a `stat` panel in the always-expanded **Block Operations** row of
`dashboards/dashboards/PostgreSQL/PostgreSQL_Overview_Extended.json`, present and unrenamed on
`percona/pmm` `main`. So this is not a dashboard regression and there is nothing to report against
the product.

The run's own Playwright trace (artifact `artifacts_@nightly`) shows it was on the page at the
moment of the check. Reading the `before`/`after` pairs out of `trace.trace`, the last four actions
of the walk are:

| t (s) | action | result |
| --- | --- | --- |
| 1534.6 | `keyboardPress PageDown` | — |
| 1534.8 | `queryCount .//button[@aria-label="Expand row"]` | `0` |
| 1536.3 | `queryCount [data-testid*="Total Blocks Operations"]` | **`1`** |
| 1536.3 | `isVisible …nth=0` | **`true`** |
| 1536.4 | `waitForSelector …nth=0` | **timeout** |

`scrollBackToPanel`'s own count found the panel and its one match was visible. 130 ms later the
wait for the *same selector* expired. Note also that no `PageUp` was pressed anywhere in that
iteration — the scan never ran, so the "walked past the panel" explanation that
`scrollBackToPanel` exists for is not what happened here.

## What actually happened: the renderer stalled

Screencast frames are emitted on paint, so their timestamps measure whether the page is painting
at all. In that trace:

```
frames: 1192 over 181.9s
largest render gaps: 8.96s, 2.50s, 2.41s, 2.33s, 2.22s, 2.21s, 2.21s, 2.15s, 2.10s, 2.04s, 1.91s, 1.76s
```

The 8.96s gap runs from 1535.0s to 1544.0s — the failing `waitForSelector` started at 1536.4s and
expired at 1541.4s, entirely inside it. Playwright polls for a selector from a script injected
*into the page*, so while Grafana's renderer is blocked that poll cannot run and the wait expires
on an element that never left the DOM. Playwright then reports the timeout as "still not present",
which is why the message names a missing panel.

Stalls of that shape are normal on this dashboard once `expandEachDashboardRow` has expanded all
ten collapsible rows and ~90 panels are mounted at once: eleven more gaps between 1.76s and 2.50s
in the same walk. A five-second budget therefore sat just above the routine case and well under the
bad one — the test was one long stall away from red on every one of its 88 iterations.

## The fix

Both metric walks gave each panel `I.waitForElement(locator, 5)`. That `5` is an explicit
narrowing of the project's own `waitForTimeout: 60000` (`pr.codecept.js:26`) down to five seconds.
Raised to `30` at both call sites.

It first landed as a named `panelWaitSeconds` constant carrying the measurement in a comment;
review dropped both in favour of the inline literal (`de7433e`, `ec9859d`), on the grounds that this
file carries almost no comments and the reasoning belongs here instead. Behaviour is identical — the
reasoning is in this body and in commit `2813934`.

**No assertion is weakened.** A panel that is genuinely absent still fails the test; it now takes
30s to do so instead of 5s, and since the walk throws on the first missing metric that costs one
wait, not 88.

## Reproduction and verification

Throwaway Linode VM (`g6-standard-6`, 6 vCPU), PMM Server pulled as
`perconalab/pmm-server:3-dev-latest` — image id `06f2a8fe42f2`, **the same image** the failing
nightly recorded to Launchable — plus `pmm-framework --database pdpgsql`, giving the
`pdpgsql_pmm_17_1_9622` service the test's `^pdpgsql_pmm_(?!.*patroni).*$` regex selects.

| Condition | Runs | Result |
| --- | --- | --- |
| idle | 1 | passed (334s) |
| 12 CPU hogs | 2 | both failed — but at `I.scrollTo`, i.e. a stall past **20s** |
| 6 CPU hogs | 1 | passed |
| **9 CPU hogs** | 4 | **1 passed, 3 failed — one of them at `I.waitForElement(…, 5)`**, the CI symptom |

That one is the reproduction that matters. On unmodified `main` (`f4f1cbc`), under 9 hogs:

```
✖ I.waitForElement([data-testid*="data-testid Panel header"][data-testid*="Total Written Files to disk"], 5)
    at Object.verifyMetricsExistencePartialMatch (./tests/pages/dashboardPage.js:1210:9)
```

Same walk, same helper, same five-second budget, a different panel — because which panel loses is
just a function of where the stall lands.

### A/B at 9 hogs, 4 runs per arm

Traces were kept for passing runs too, via an **untracked** `probe.codecept.js` copy with
`keepTraceForPassedTests: true` — the tracked config was not edited.

| Arm | commit | pass | fail at `waitForElement(…, 5)` | fail at `I.scrollTo` (20s action timeout) |
| --- | --- | --- | --- | --- |
| baseline | `f4f1cbc` | 1 | **1** | 2 |
| fixed | `2813934` | 1 | **0** | 3 |

**Read the failing step, not the pass count.** The pass counts are equal and mean little: this box
under 9 hogs stalls far harder than any GitHub runner has — max gaps of 21–70s per run against
CI's 8.96s — so once the 5s budget is out of the way the 20s `scrollTo` becomes the binding
constraint and the arms converge. What the fix is supposed to do is remove the exposure at the
panel wait, and it does: zero failures there across four runs, against one in four on baseline.

The clearest single data point is fixed run 3, which **passed** while its own trace recorded
**11 render gaps over 5s** (four over 20s, longest 26.8s):

```
FIX-3: frames=1162 max=26.80s p99=4.93s gaps>5s=11 gaps>20s=4
```

Per-run stall measurements (screencast gaps):

```
baseline-1: max=25.69s gaps>5s=8   baseline-2: max=70.32s gaps>5s=9
baseline-3: max=28.01s gaps>5s=2   baseline-4: max=28.08s gaps>5s=9
FIX-1: max=26.68s gaps>5s=3        FIX-2: max=23.62s gaps>5s=9
FIX-3: max=26.80s gaps>5s=11       FIX-4: max=21.24s gaps>5s=1
```

`npx eslint tests/pages/dashboardPage.js` — clean on `2813934`. The review commits that followed only
replaced the constant with its own value, so the verified behaviour is unchanged on the current head.

### CI on this branch

`E2E tests Matrix` is **green on the current head `ec9859d`** — all 37 checks success or skipped,
none failing. It does not select `@nightly` or `@dashboards`, so PMM-T2048 itself did not run — but
one of the two changed call sites did:

> **FB E2E tests / PGSM UI tests / e2e tests: `@pgsm-pmm-integration`** — job
> [103914626332](https://github.com/percona/pmm-qa/actions/runs/34824860457/job/103914626332),
> conclusion `success`, completed 2026-09-14T09:19:31Z. Its only selected file,
> `tests/qa-integration/pmm_pgsm_integration_test.js`, calls `verifyMetricsExistence` twice against
> `postgresqlInstanceSummaryDashboard`.
>
> The same job on the pre-review head `2813934` —
> [103664570502](https://github.com/percona/pmm-qa/actions/runs/34734944872/job/103664570502),
> `success`, completed 2026-09-13T03:41:25Z — recorded Launchable **19 found, 19 passed, 0 failed**
> over 21.25 min, gate `PASSED`, build `sha256:06f2a8fe42f2…`, the same PMM image as the failing
> nightly.

So the changed helper walks a real dashboard to completion in CI with the 30s budget in place. That
is a no-regression result, not a confirmation that PMM-T2048 is fixed.

### What this does not prove

- **`SMOKE TESTED`, not `VERIFIED`, on the end-to-end claim.** The nightly's PMM Server is external
  and already destroyed, and `nightly-e2e-tests-matrix.yml` cannot be re-run against it. The
  branch's own CI exercises the changed helper (above) but not PMM-T2048's tags, so the only thing
  that can confirm the 5s budget was PMM-T2048's last remaining flake is a subsequent nightly. The Jenkins dispatcher does take `PMM_QA_GIT_BRANCH`, so this branch can be put through
  a real nightly ahead of merge if someone wants that.
- **The 30s figure is sized from one CI observation** (8.96s) plus the local distribution, not from
  a long baseline of nightly traces. It is ~3.3× the worst stall CI has shown. If a future nightly
  fails at this wait again, the budget is the thing to revisit, not the walk.

## Left out on purpose

Three of the three local failures with the fix applied are `I.scrollTo` timing out on
`locator('#grafana-iframe').contentFrame().locator(':root')` after 20s — the global
`timeout: 20000` in `pr.codecept.js`. It is the same class of defect (a render stall starving a
wait), and raising it would very likely help too.

It is **not** changed here, for two reasons: no CI run has ever been observed failing there — the
20s exposure only appeared on a box loaded well past what a GitHub runner does — and that value is
the action timeout for *every* test in the CodeceptJS suite, so moving it is a change to the whole
suite rather than to this walk. Flagging it as a real risk rather than folding it in.

Worth a separate look by someone who owns this helper: the walk's own design is what produces the
stalls it then times out on. `expandEachDashboardRow` mounts every panel of a ten-row dashboard
before the walk starts, and this dashboard has ~90. Rendering fewer at a time would remove the
stalls rather than budget for them — but that is a redesign of a helper shared across the dashboard
suites, not a fix for this failure.

https://claude.ai/code/session_01YBPnBQ1TTQWJtQ7dsM8jAY

---
_Generated by [Claude Code](https://claude.ai/code/session_01YBPnBQ1TTQWJtQ7dsM8jAY)_

---

### DIFF
--- a/codeceptjs-e2e/tests/pages/dashboardPage.js
+++ b/codeceptjs-e2e/tests/pages/dashboardPage.js
@@ -27,6 +27,12 @@ const ValkeyPersistenceDetailsDashboard = require('../pages/dashboards/valkey/va
 const ValkeyReplicationDashboard = require('../pages/dashboards/valkey/valkeyReplicationDashboard');
 const ValkeySlowlogDashboard = require('../pages/dashboards/valkey/valkeySlowlogDashboard');
 
+// Re-rendering an expanded dashboard blocks Grafana's renderer for seconds at a time, and a
+// blocked renderer cannot answer Playwright's in-page poll, so a panel that is on screen reports
+// as "still not present". Nightly run 34727137335 stalled 2.0-2.5s repeatedly during one metric
+// walk and once for 8.96s, which is what a five-second budget lost to.
+const panelWaitSeconds = 30;
+
 module.exports = {
   // insert your locators and methods here
   // setting locators
@@ -1197,7 +1203,7 @@ module.exports = {
       I.pressKey('PageDown');
       await this.expandEachDashboardRow();
       await this.scrollBackToPanel(this.graphsLocator(metrics[i]));
-      I.waitForElement(this.graphsLocator(metrics[i]), 5);
+      I.waitForElement(this.graphsLocator(metrics[i]), panelWaitSeconds);
       I.scrollTo(this.graphsLocator(metrics[i]));
     }
   },
@@ -1207,7 +1213,7 @@ module.exports = {
       I.pressKey('PageDown');
       await this.expandEachDashboardRow();
       await this.scrollBackToPanel(this.graphsLocatorPartialMatch(metrics[i]));
-      I.waitForElement(this.graphsLocatorPartialMatch(metrics[i]), 5);
+      I.waitForElement(this.graphsLocatorPartialMatch(metrics[i]), panelWaitSeconds);
       I.scrollTo(this.graphsLocatorPartialMatch(metrics[i]));
     }
   },


codeceptjs-e2e/tests/pages/dashboardPage.js
```diff
@@ -27,6 +27,12 @@ const ValkeyPersistenceDetailsDashboard = require('../pages/dashboards/valkey/va
 const ValkeyReplicationDashboard = require('../pages/dashboards/valkey/valkeyReplicationDashboard');
 const ValkeySlowlogDashboard = require('../pages/dashboards/valkey/valkeySlowlogDashboard');
 
+// Re-rendering an expanded dashboard blocks Grafana's renderer for seconds at a time, and a
+// blocked renderer cannot answer Playwright's in-page poll, so a panel that is on screen reports
+// as "still not present". Nightly run 34727137335 stalled 2.0-2.5s repeatedly during one metric
+// walk and once for 8.96s, which is what a five-second budget lost to.
+const panelWaitSeconds = 30;
```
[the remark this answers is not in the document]
travagliad: Fair on density — the file carries almost no comments and the stall distribution belongs in the PR body. Optional, so it doesn't start a push on its own; the two-line form rides this PR's next code push if one comes, and a reviewer can apply the suggestion directly otherwise. Keeping the bounded const rather than inheriting `waitForTimeout: 60000`: the explicit budget is what carries the measured floor, and it bounds the cost of a genuinely missing panel to 30s instead of 60s.

---

travagliad:
```suggestion
```

codeceptjs-e2e/tests/pages/dashboardPage.js
```diff
@@ -1197,7 +1203,7 @@ module.exports = {
       I.pressKey('PageDown');
       await this.expandEachDashboardRow();
       await this.scrollBackToPanel(this.graphsLocator(metrics[i]));
-      I.waitForElement(this.graphsLocator(metrics[i]), 5);
+      I.waitForElement(this.graphsLocator(metrics[i]), panelWaitSeconds);
```
travagliad:
```suggestion
      I.waitForElement(this.graphsLocator(metrics[i]), 30);
```

codeceptjs-e2e/tests/pages/dashboardPage.js
```diff
@@ -1207,7 +1213,7 @@ module.exports = {
       I.pressKey('PageDown');
       await this.expandEachDashboardRow();
       await this.scrollBackToPanel(this.graphsLocatorPartialMatch(metrics[i]));
-      I.waitForElement(this.graphsLocatorPartialMatch(metrics[i]), 5);
+      I.waitForElement(this.graphsLocatorPartialMatch(metrics[i]), panelWaitSeconds);
```
travagliad:
```suggestion
      I.waitForElement(this.graphsLocatorPartialMatch(metrics[i]), 30);
```

codeceptjs-e2e/tests/pages/dashboardPage.js
```diff
@@ -27,6 +27,7 @@ const ValkeyPersistenceDetailsDashboard = require('../pages/dashboards/valkey/va
 const ValkeyReplicationDashboard = require('../pages/dashboards/valkey/valkeyReplicationDashboard');
 const ValkeySlowlogDashboard = require('../pages/dashboards/valkey/valkeySlowlogDashboard');
 
+
```
travagliad:
```suggestion
```

yurkovychv (approved):

Nailya (approved):

PR #1437, opened 2026-09-15
migrate(dashboards): verifyPMMHealthDashboard CodeceptJS to Playwright

kiranvuyurru:
Migrates `codeceptjs-e2e/tests/dashboards/verifyPMMHealthDashboard_test.js` to `e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts`.

**Scenarios:** PMM-T1988 (1 row) - tags `@nightly @gssapi-nightly` preserved.
**Setup:** PMM Server only, no monitored database; local Docker via provisioning/.
**Coverage:** already selected - `test_execution_playwright` in `nightly-e2e-tests-matrix.yml` greps `@nightly`, so no workflow change (`expected_test_jobs` stays 3). `@gssapi-nightly` has no workflow consumer. Retirement does drop the scenario from `codeceptjs-e2e/package.json`'s catch-all `e2e` script, whose only caller is cross-repository.
**Retired:** `verifyPMMHealthDashboard_test.js` -> `_migrated.js`; no job deleted - the CodeceptJS `@nightly` entry keeps 50 scenarios across 12 files.
**Deviations:** the source's `//section[contains(@data-testid,"<name>")]//div[contains(@data-testid,"panel content")]` is re-derived as `builders.panelByExactName(name).getByTestId('data-testid panel content')`, both resolving to count 1 per panel on live DOM; `verifyStatuses()` is not ported as a POM method, its assertions live in the test body.
**Run:** https://github.com/percona/pmm-qa/actions/runs/35030660505


kiranvuyurru:
GitHub Actions: https://github.com/percona/pmm-qa/actions/runs/35030660505

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -0,0 +1,33 @@
...
+});
+
+pmmTest(
+  'PMM-T1988 - Verify that all statuses on PMM Health dashboards are UP @nightly  @gssapi-nightly',
+  async ({ dashboard, page }) => {
+    await page.goto(dashboard.pmmHealth.url);
+
+    for (const panelName of statusPanels) {
+      const panelContent = dashboard.builders
+        .panelByExactName(panelName)
```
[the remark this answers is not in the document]
kiranvuyurru: Not reproducible against a live PMM Server, so no change made. `graph/d/pmm-health` renders 17 `section[data-testid]` panels and each of the seven asserted titles resolves to **count 1** — verified through Playwright MCP in two independent review passes, and the source XPath `//section[contains(@data-testid,"<name>")]` resolves to 1 for the same seven. Strict mode is also what proves it: duplicates would throw `strict mode violation: resolved to N elements`, and this test passed twice (local control worktree 8.0s, publish worktree 9.6s), including on `ManageD Status`, the first panel in the loop. If the duplicates you found come from a dashboard JSON, they are likely panels inside a collapsed row, which never render and so never enter the locator's match set.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -0,0 +1,33 @@
...
+        .getByTestId('data-testid panel content');
+
+      await expect(panelContent).toBeVisible({ timeout: Timeouts.ONE_MINUTE });
+      await expect(panelContent).toContainText('UP', { timeout: Timeouts.TEN_SECONDS });
+    }
+  },
+);
```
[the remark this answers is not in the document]
kiranvuyurru: Declining, because it breaks scenario-count fidelity — the one invariant a migration may not trade. The source is a bare `Scenario(...)` whose loop lives in the page object (`verifyStatuses()`), not a `Data(...)` scenario, so it is one scenario and must port to one test; splitting makes it seven, and seven dashboard loads against the shared Jenkins-managed nightly server instead of one.

The precedent does not transfer: `homeDashboards.test.ts:25` loops because **its** source is `Data(panels).Scenario(...)` with seven rows (`verifyHomeDashboards_migrated.js:20`), so one-test-per-panel there is the rule being followed — "one test per data row", title suffixed `| <that one value>` — not a house style for loops.

Your diagnostic point is real and unaddressed by either version: the first failing panel leaves the other six unmeasured. That is equally true of the CodeceptJS source, so the port preserves it rather than introducing it. Playwright does name the culprit — the panel title is inside the failing locator's test-id — but converting this to per-panel coverage is a coverage change and belongs in its own PR, not in the migration.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/dashboards.page.ts
```diff
@@ -30,6 +30,7 @@ export default class Dashboards extends BasePage {
   readonly mongo: MongoDashboardsType = MongoDashboards;
   readonly mysql: MysqlDashboardsType = MysqlDashboards;
   readonly os: OperatingSystemDashboardsType = OperatingSystemDashboards;
+  readonly pmmHealth = { url: 'graph/d/pmm-health/pmm-health' };
```
travagliad: AI got crazy 😆 

Claude, always create the page object for each different page we have, including this one, move panels info to that file instead of putting it all into tests too, check what other tests that checks all panels does, I think valkey is a good example but maybe there is others.

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  'Node Status',
+  'Clickhouse Status',
+] as const;
+
+pmmTest.beforeEach(async ({ grafanaHelper }) => {
+  await grafanaHelper.authorize();
+});
+
+pmmTest(
+  'PMM-T1988 - Verify that all statuses on PMM Health dashboards are UP @nightly  @gssapi-nightly',
```
travagliad, line 10: We need gssapi-nightly runner for playwright, or we will effectively skip the test there

e2e_tests/pages/dashboards/dashboards.page.ts
kiranvuyurru: Done in `01fd00d8`. Added `e2e_tests/pages/dashboards/pmmHealth.ts` implementing `DashboardInterface` the way `pages/dashboards/valkey/*.ts` do — `url`, a `GrafanaPanel[]` of the seven status panels, empty `noDataMetrics` — and `dashboards.page.ts` now registers `new PmmHealthDashboard()` instead of the inline `{ url }` object, matching how `home` registers `HomeDashboard`. The test drops its `statusPanels` array and iterates `dashboard.pmmHealth.metrics`.

I did not adopt valkey's `dashboard.verifyMetricsPresent(...)` call: it checks that panels are present and have data, not that the content reads `UP`, which is the assertion PMM-T1988 exists for. So the panel data moved to the page object as you asked, and the two per-panel assertions stayed in the test.

Verified locally against a server-only PMM: lint and `tsc --noEmit` clean, test green.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
kiranvuyurru: Agreed, and it now exists — @kiranvuyurru added a Playwright runner greping bare `@gssapi-nightly` to the Jenkins pipeline. Commit `9d5b7648` on this branch records that, because the retirement commit body had claimed the tag "is not a selection route and never was", which is no longer true; this repo squash-merges commit messages, so the claim and its retraction land on `main` together.

No workflow change is owed in this repo: the consumer is the Jenkins pipeline, and the migrated test carries `@gssapi-nightly` verbatim from the source, so that runner selects it with no edit here. It is also selected independently by `nightly-e2e-tests-matrix.yml:153` `test_execution_playwright` (`tags_for_tests: '@nightly'`), so the scenario was never skipped outright — but you are right that before the Jenkins runner the GSSAPI tag contributed nothing to selection.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/pmmHealth.ts
```diff
@@ -0,0 +1,16 @@
...
+export default class PmmHealthDashboard implements DashboardInterface {
+  url = 'graph/d/pmm-health/pmm-health';
+  metrics: GrafanaPanel[] = [
+    { name: 'ManageD Status', type: 'unknown' },
+    { name: 'VictoriaMetrics Status', type: 'unknown' },
+    { name: 'PostgreSQL Status', type: 'unknown' },
+    { name: 'QAN API Status', type: 'unknown' },
+    { name: 'Grafana Status', type: 'unknown' },
+    { name: 'Node Status', type: 'unknown' },
+    { name: 'Clickhouse Status', type: 'unknown' },
```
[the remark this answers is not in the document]
travagliad: @kiranvuyurru  This seems accurate, can you review if it works as intended on linode VM?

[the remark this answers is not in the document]
kiranvuyurru: You were right and my two earlier answers were wrong. Fixed in `288f329c`.

I refuted this twice with a live count of 1 per title plus the green runs, and explained your JSON duplicates as collapsed rows. Both halves were wrong. Measured on a live server: the dashboard has **89 grid items and zero collapsed rows**, and a default load renders only **19** panel sections — Grafana lazy-renders below the fold, so my count of 1 was an artifact of the viewport, not evidence of uniqueness. Driving `loadAllPanels()` to scroll every grid item into view renders **81** sections, and `ManageD Status`, `QAN API Status`, `Grafana Status` and `Clickhouse Status` then resolve to **2 each** — exactly the pairs you named (1086/1045, 1085/1071). The test was passing for a reason a taller viewport or a faster render could remove at any time.

Each summary panel is now pinned by its Grafana panel id via a new `gridItemByPanelId` builder scoping to `[data-griditem-key="grid-item-<id>"]` and then to the title, so id and title must both match. Verified with all 81 sections rendered: each of the seven resolves to exactly one grid item, header and content reading `UP`, all at y=38px; the duplicates sit at y=1710px and below.

One correction for the learning recorded on the withdrawn thread — it says these duplicates belong to collapsed rows and are not rendered. They are not in collapsed rows, and they do render once scrolled to.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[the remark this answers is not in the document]
kiranvuyurru: @travagliad — please do not rely on this answer; it is repeating a wrong explanation that I gave CodeRabbit earlier, and I have since disproved it.

**The duplicate titles are not in collapsed rows.** Measured on a live PMM Server: the dashboard has 89 grid items and **zero** collapsed rows. A default page load renders only **19** `section[data-testid]` panels, which is why every title appeared to resolve to 1 — Grafana lazy-renders below the fold, so that count was an artifact of the viewport, not proof of uniqueness. After `loadAllPanels()` scrolls every grid item into view, **81** sections render and `ManageD Status`, `QAN API Status`, `Grafana Status` and `Clickhouse Status` resolve to **2 each**.

So CodeRabbit's original finding was correct, my two refutations of it were wrong, and it withdrew on my say-so. The strict-mode failure was real but latent — the test passed only because the lower duplicates had not rendered yet.

Fixed in `288f329c`: each summary panel is pinned by its Grafana panel id via a `gridItemByPanelId` builder scoping to `[data-griditem-key="grid-item-<id>"]` and then to the title, so id and title must both match. With all 81 sections rendered, each of the seven resolves to exactly one grid item, header and content reading `UP` at y=38px; the duplicates sit at y=1710px and below.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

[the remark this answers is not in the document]
kiranvuyurru: Worth flagging before you review: the version you called accurate had a real defect, now fixed in `288f329c`, so the panel list on this line has changed since your comment.

What is verified, on a local server-only Docker PMM (`provisioning/setup.ts`, `perconalab/pmm-server:3-dev-latest`):

- 89 grid items, **zero** collapsed rows on `graph/d/pmm-health`.
- A default load renders only 19 panel sections; after `loadAllPanels()` scrolls everything into view, 81 render and four of the seven titles resolve to **2** elements. The earlier version would have hit a Playwright strict-mode violation as soon as a duplicate rendered.
- With all 81 sections rendered, each summary panel now resolves to exactly one grid item, header and content reading `UP`, pinned by Grafana panel id (`[data-griditem-key="grid-item-<id>"]`) plus the title.
- Test green twice, lint and `tsc --noEmit` clean, migration convention script exit 0.

What is **not** verified: a Linode VM run. I cannot provision one from this session, so I have not done it and would rather say so than imply coverage that does not exist. The panel ids are read from the running Grafana, so the check worth making there is simply that the seven still resolve 1:1 — if PMM ships a dashboard whose panel ids differ, this fails loudly with a missing locator rather than silently asserting the wrong panel, which is why ids were chosen over `.first()`.

Leaving this thread open for @kiranvuyurru rather than resolving it, since the Linode check is yours to make.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e_tests/pages/dashboards/pmmHealth.ts
```diff
@@ -3,14 +3,15 @@ import DashboardInterface from '@interfaces/dashboard';
...
+  statusPanels = [
+    { name: 'ManageD Status', panelId: '1086' },
+    { name: 'VictoriaMetrics Status', panelId: '1088' },
+    { name: 'PostgreSQL Status', panelId: '1089' },
+    { name: 'QAN API Status', panelId: '1085' },
+    { name: 'Grafana Status', panelId: '1087' },
+    { name: 'Node Status', panelId: '1090' },
+    { name: 'Clickhouse Status', panelId: '1091' },
   ];
+  metrics: GrafanaPanel[] = this.statusPanels.map(({ name }) => ({ name, type: 'unknown' as const }));
```
travagliad: This is not the approach we agreed on when defining how to define the metrics (panels) to see, the previous was the most accurate one, let's switch back to that, if panelId is strictly necessary let's do it without breaking the contract

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -11,9 +11,10 @@ pmmTest(
   async ({ dashboard, page }) => {
     await page.goto(dashboard.pmmHealth.url);
 
-    for (const panel of dashboard.pmmHealth.metrics) {
+    for (const panel of dashboard.pmmHealth.statusPanels) {
       const panelContent = dashboard.builders
-        .panelByExactName(panel.name)
+        .gridItemByPanelId(panel.panelId)
+        .getByTestId(`data-testid Panel header ${panel.name}`)
         .getByTestId('data-testid panel content');
```
travagliad: raw methods from playwright shouldn't be used on the test file, we should move that to page object

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      const panelContent = dashboard.builders
+        .panelByExactName(panel.name)
+        .getByTestId('data-testid panel content');
+
+      await page.goto(dashboard.pmmHealth.url);
+      await expect(panelContent).toBeVisible({ timeout: Timeouts.ONE_MINUTE });
+      await expect(panelContent).toContainText('UP', { timeout: Timeouts.TEN_SECONDS });
+    },
+  );
+}
```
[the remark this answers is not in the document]
kiranvuyurru: already added

e2e_tests/tests/dashboards/pmmHealthDashboard.test.ts
```diff
@@ -1,24 +1,23 @@
...
-  'PMM-T1988 - Verify that all statuses on PMM Health dashboards are UP @nightly  @gssapi-nightly',
-  async ({ dashboard, page }) => {
-    await page.goto(dashboard.pmmHealth.url);
-
-    for (const panel of dashboard.pmmHealth.statusPanels) {
-      const panelContent = dashboard.builders
-        .gridItemByPanelId(panel.panelId)
-        .getByTestId(`data-testid Panel header ${panel.name}`)
-        .getByTestId('data-testid panel content');
+for (const panel of new PmmHealthDashboard().metrics) {
```
travagliad: We had a better way before, why this change? The agent had already refused to do it, reason being, there is no need to start a new page, navigation, fixtures, for repeating tests on same page.

kiranvuyurru: done

travagliad: Where? 😕

travagliad (approved):

PR #1442, opened 2026-09-16
Refine agent and skill documentation

kiranvuyurru:


.claude/scripts/instruction-budget.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
travagliad, line 1: Is hooks directory the best for this? Looks like a good fit for scripts as this is meant for stuff that we put under claude hooks

.claude/agents/skill-gardener-publisher.md
travagliad, line 21: Can we add this?

```suggestion

**If the lesson is contradictory to any existing behavior, edit the existing behavior instead of only adding text.**  Avoid incremental changes whenever it's possible, prefer editing of existing and broken knowledge, if such is not possible, then you increment, preferably into new lines.

```

.claude/scripts/instruction-budget.sh
kiranvuyurru: moved to scripts

PR #1443, opened 2026-09-16
migrate(dashboards): verifySearchDashboards CodeceptJS to Playwright

kiranvuyurru:
Ports `codeceptjs-e2e/tests/dashboards/verifySearchDashboards_test.js` to `e2e_tests/tests/dashboards/searchDashboards.test.ts` and retires the source.

**Scope:** PMM-T1091 plus the PMM-T1086 `Data` table — 10 tests over 9 folders and 48 dashboard names, one-for-one with the CodeceptJS dry-run. Tag strings copied verbatim, interior double space included.

**Run:** no PR-CI job selects `@nightly` — it lives only in the `workflow_dispatch`-only nightly matrix. Executed locally against a server-only PMM 3.x (`provisioning/setup.ts`): **10 passed, 1.0m**, with retries at the config default of 0 locally, so nothing was masked. For CI, dispatch Jenkins `pmm3-ui-tests-nightly-gha` with `PMM_QA_GIT_BRANCH=migrate-dashboards-verifySearchDashboards`.

**Notes for review**

- `builders.dashboardRow` filters the row test id on `icon-apps`. Grafana emits the same `data-testid browse dashboards row <title>` for a folder row and a dashboard row, and the `PMM Health` folder contains a dashboard also titled `PMM Health` — so the unfiltered locator resolves to two nodes and strict mode fails where CodeceptJS's non-strict `seeElementInDOM`/`see` did not. The filter's inner locator is rooted at `page`: `filter({ has })` matches it relative to each outer match, so re-entering `frameLocator` there is a no-op.
- The MySQL folder is absent from the `Data` table, matching the source's own `return`. It stays in PMM-T1091's expected folder list because the source's T1091 maps the page object's whole `folders` object, not the filtered table.
- The folder inventory lives in `e2e_tests/testdata/dashboards.registry.ts` as `FOLDER_DASHBOARDS`, beside the `DASHBOARDS` list that file already exports. The two are deliberately not merged: `DASHBOARDS` covers 7 folders to this list's 9, differs in membership within shared folders, and keys by URL slug where the browse page renders display titles. `tsconfig.json` gains `@testdata/*` so the import is aliased like every other.
- The source's `Feature.retry(1)` is dropped rather than pinned — `playwright.config.ts` already sets `CI ? 2 : 0`, and pinning 1 would make CI less tolerant and local runs more tolerant than that policy.
- `codeceptjs-e2e/tests/pages/dashboards/components/searchDashboardsModal.js` is **not** retired: `tests/upgrade/dashboards_test.js` still injects it for its own `@post-dashboards-upgrade` scenarios.
- No workflow change. Both titles keep `@nightly`, which `nightly-e2e-tests-matrix.yml` already greps on `test_execution_playwright`, so `expected_test_jobs` stays 3 and retirement orphans no job.


e2e_tests/README.md
```diff
@@ -112,6 +112,7 @@ For test, page-object, fixture, and helper conventions, see [CONTRIBUTING.md](./
 - `@pmm-psmdb-integration`
 - `@pmm-valkey-integration`
 - `@post-release`
+- `@post-upgrade`
```
travagliad: Didn't we dropped this?

kiranvuyurru: yes, let me remove it

e2e_tests/tests/dashboards/searchDashboards.test.ts
```diff
@@ -0,0 +1,48 @@
...
+          await expect(async () => {
+            if (!(await row.isVisible())) await page.keyboard.press('PageDown');
+
+            await expect(row).toContainText(dashboardName, { timeout: Timeouts.ONE_SECOND });
+          }).toPass({ intervals: [Timeouts.ONE_SECOND], timeout: Timeouts.TEN_SECONDS });
+        }
+      });
+    },
+  );
+}
```
travagliad: Can't the for loop live under a single test that checks all expands? Don't see the reason to close and reopen the same page multiple times to perform a check that lives here

kiranvuyurru: agreed

travagliad (approved):

PR #1446, opened 2026-09-17
migrate(dashboards): verifyNodesOverviewDashboard CodeceptJS to Playwright

kiranvuyurru:
Migrates `codeceptjs-e2e/tests/dashboards/verifyNodesOverviewDashboard_test.js` to `e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts`.

**Scenarios:** PMM-T1642 (1 scenario) - tag `@docker-configuration` preserved.
**Setup:** `node provisioning/setup.ts --db pdpgsql` (matching the job's `setup_services`) plus a pmm-client node labelled `environment=dev`; local Docker via provisioning/.
**Coverage:** no workflow change - `docker_configuration_playwright` in `fb-e2e-suite.yml` already greps `@docker-configuration` and installs the host client, so it selects this test as-is. The job's `setup_services` is `--database pdpgsql`, where the CodeceptJS `docker_configuration` job used `-h`; neither CI failure below traces to that difference.
**Retired:** `verifyNodesOverviewDashboard_test.js` -> `_migrated.js`; no job deleted, `docker_configuration` keeps 10 tag occurrences across 4 files.
**Deviations:** `I.wait(60)` becomes a poll on `node_boot_time_seconds{environment="dev"}` at a 2m budget; `verifyThereAreNoGraphsWithoutData(4)` becomes `dashboard.verifyAllPanelsHaveData(dashboard.os.nodesOverview.noDataMetrics)`, asserting the empty panels by name instead of against a count budget (see below); `--force` added to the `pmm-admin config` call (see below); dropped as inert - two unread consts, a focus click, the End-key expand loop that `loadAllPanels` covers, and a `24`/`hour` title exemption that matches none of the 81 panel names.

**Assertion change (review feedback, [thread](https://github.com/percona/pmm-qa/pull/1446#discussion_r4048747835)):** at review request the ported count budget was replaced with the existing `verifyAllPanelsHaveData` allow-list helper, and `NodesOverviewDashboard.noDataMetrics` populated with `['DB Instances', 'Min DB Uptime', 'Top Disk Latency', 'Top 5 Disk Latency']`. The two disk-latency panels came from the first CI run (`Metrics without data are: Top Disk Latency,Top 5 Disk Latency`): the GitHub runner's virtual disk exposes no latency series, unlike the container node the list was first measured on, and four is the budget the CodeceptJS scenario passed. The first two are the only panels on the dashboard sourced purely from database exporters (`mysql_global_status_uptime` / `mongodb_*` / `pg_up` / `postgresql_up`, and `pg_postmaster_start_time_seconds`); every other panel is node_exporter backed. The test labels a node that runs no database - `pmm-admin config --custom-labels=environment=dev` labels the test-runner host, while pdpgsql lives on its own unlabeled node - so filtering to `environment=dev` selects a node-only node and exactly those two are empty by construction.

This supersedes the earlier 3-vs-4 budget question: the literal `4` is gone, and `acceptableNoDataCount = 3` is not read by this test.

It improves diagnostics without closing the filter gap raised earlier in review: a silently ignored `var-environment` renders the dashboard unfiltered, zero panels are empty, and the allow-list passes exactly as the count budget did. It catches the label failing to reach the metrics, and now names what is missing when it does.

**Local verification** (not a substitute for CI - `cliHelper.execute` runs on the test-runner host, so `sudo pmm-admin config` cannot run on Windows; the navigation and assertion were driven directly and that first line is unexercised locally, against a pmm-client *container* node rather than CI's generic runner node):
- three consecutive loads of the filtered dashboard: `["DB Instances","Min DB Uptime"]` every time, `scrolled=2 attached=2 visible=2`
- the assertion reading the list from the POM: passes
- negative case, an environment value matching no node: fails with `Metrics without data are: …`, naming all 58 panels

**`--force` (fix on `df0890f4`):** the first CI run's two retries died on the test's first line - `Failed to register pmm-agent on PMM Server: Node with name runnervm7g52i already exists. If you want override node, use --force option.` ([run](https://github.com/percona/pmm-qa/actions/runs/35374900403/job/105697596890), `1 failed, 6 passed`). Attempt 1's `pmm-admin config` succeeded and registered the host (`Registered.`, then the run failed later at the assertion - the disk-latency panels above). Playwright's CI `retries: 2` reran the test into the same environment, and a second registration under the same hostname is refused: the rerun collision `codeceptjs-migration/branch-workflow.md` documents. `--force` makes the command idempotent across retries; every other `pmm-admin config` call in this repo already passes it. The CodeceptJS scenario passed on main on its first attempt ([run](https://github.com/percona/pmm-qa/actions/runs/35299059686)), so it never retried into this.

**Run:** PR CI job `docker_configuration_playwright` (`e2e-tests-matrix.yml` -> `fb-e2e-suite.yml`) selects PMM-T1642. The run on `68958103` is in progress; it will be linked here on completion.


kiranvuyurru:
GitHub Actions: https://github.com/percona/pmm-qa/actions/runs/35275144081

Job `FB E2E tests / Docker configuration tests / e2e tests: @docker-configuration` (the `docker_configuration_playwright` caller) selects PMM-T1642 on this branch.

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    );
+    await dashboard.loadAllPanels();
+    await expect
+      .poll(() => dashboard.elements.noDataPanelName.count(), {
+        message: `Nodes Overview filtered by environment=${environment} should show at most 4 panels without data`,
+        timeout: Timeouts.ONE_MINUTE,
+      })
+      .toBeLessThanOrEqual(4);
+  },
+);
```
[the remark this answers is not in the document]
kiranvuyurru: Correct, and inherited rather than introduced: the CodeceptJS scenario asserts the same `verifyThereAreNoGraphsWithoutData(4)` on the same `environment=dev` URL, so an ignored filter passes there too (unfiltered measures 0 no-data panels against a budget of 4).
Adding a node-list assertion would be a coverage change, which this migration explicitly forbids, so the port stays faithful.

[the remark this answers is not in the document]
kiranvuyurru: Saying so explicitly: the gap is real and pre-existing. Measured live - unfiltered 0 no-data panels, `environment=dev` with the label 2, `environment=dev` without it 60, budget 4 - so the assertion catches the label failing to propagate but not a silently ignored filter, exactly as the source does.
That makes a node-list check a coverage addition rather than a port fix, so I am leaving it and resolving this.

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        to: 'now',
+      }),
+    );
+    await dashboard.loadAllPanels();
+    await expect
+      .poll(() => dashboard.elements.noDataPanelName.count(), {
+        message: `Nodes Overview filtered by environment=${environment} should show at most 4 panels without data`,
+        timeout: Timeouts.ONE_MINUTE,
+      })
+      .toBeLessThanOrEqual(4);
```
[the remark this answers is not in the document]
kiranvuyurru: Measured rather than assumed: on this dashboard attached and visible counts are identical in every state the test reaches - 60/60 on the unlabelled `dev` URL and 2/2 once the label lands, identical title sets, and no unmounting between a top and bottom scroll - so `.count()` is not stricter here in practice.
The thinner failure message is a fair point and changes no verdict; I am not re-opening a branch that has already cleared its final review for a diagnostics-only edit.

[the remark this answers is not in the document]
kiranvuyurru: Confirmed, and the 4 stays. One correction to the framing though: this is not filtered versus unfiltered - the 4 is the literal the source scenario passes (`verifyNodesOverviewDashboard_test.js:23`) while the POM 3 belongs to a different scenario whose only CodeceptJS consumer is already retired.
Two scenarios, two budgets, each faithful to its own source; recording it here rather than in the PR body since neither governs the other.

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
```diff
@@ -0,0 +1,34 @@
+import pmmTest from '@fixtures/pmmTest';
+import { expect } from '@playwright/test';
+import { Timeouts } from '@helpers/timeouts';
+
+const environment = 'dev';
```
travagliad: This config is only for the test that uses it, no reason to make It a top-level configuration that other tests in future might want to use. Just insert under the test that uses it

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
travagliad: Please confirm, I believe we have methods in place that does this same validation created by Peter

kiranvuyurru: Confirmed that a shared method exists, but it does not perform this validation.

`DashboardsPage.verifyAllPanelsHaveData(noDataMetrics)` (`e2e_tests/pages/dashboards/dashboards.page.ts:210`, 15 call sites across 6 test files) takes an **allow-list of panel names** that are permitted to be empty and fails if any panel outside that list has no data. `NodesOverviewDashboard.noDataMetrics` is `[]`, so calling it here would require every panel on the dashboard to have data.

This test asserts a **count budget** instead - at most 4 empty panels - which is the literal the CodeceptJS source passes (`verifyThereAreNoGraphsWithoutData(4)` in `verifyNodesOverviewDashboard_test.js:23`). Measured live on this branch: `environment=dev` settles at 2 empty panels, so the budget is doing real work; the `dev` node genuinely has a few panels with no data, and an allow-list of `[]` would red the run.

So it is not a drop-in swap - it would tighten the assertion and change what the test covers relative to the scenario being ported. The closest existing count-budget code is inline in `homeDashboards.test.ts:89-99` against `acceptableNoDataCount`, not a shared method.

Happy to go either way:
1. Keep the faithful port as is (my preference for this migration PR).
2. Switch to `verifyAllPanelsHaveData` and populate `NodesOverviewDashboard.noDataMetrics` with the panels that are legitimately empty - stricter and better diagnostics, but a coverage change I would want a green run behind.

Which would you prefer?

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
kiranvuyurru: Done in 48127d45 - `environment` now lives inside the test body.

e2e_tests/tests/dashboards/nodesOverviewDashboard.test.ts
travagliad: > Switch to verifyAllPanelsHaveData and populate NodesOverviewDashboard.noDataMetrics with the panels that are legitimately empty - stricter and better diagnostics, but a coverage change I would want a green run behind.

kiranvuyurru: Done in e7613bcf.

`NodesOverviewDashboard.noDataMetrics` is now `['DB Instances', 'Min DB Uptime']` and the test calls `dashboard.verifyAllPanelsHaveData(dashboard.os.nodesOverview.noDataMetrics)`. The literal `4`, the inline `.count()` and the now-unused `expect` import are gone.

**Why those two.** They are the only panels on the dashboard sourced purely from database exporters:

- `DB Instances` - `mysql_global_status_uptime … or mongodb_mongos_instance_uptime_seconds … or pg_up … or postgresql_up`
- `Min DB Uptime` - `mysql_global_status_uptime … or mongodb_instance_uptime_seconds … or (time() - pg_postmaster_start_time_seconds)`

Everything else is node_exporter backed. The test labels a node that runs no database - in CI `pmm-admin config --custom-labels=environment=dev` labels the test-runner host, while the pdpgsql database lives on its own unlabeled container node - so filtering to `environment=dev` selects a node-only node and exactly those two panels are empty by construction. The neighbouring `Nodes` panel survives because its query ends in `or vector(0)`; `DB Instances` has no such fallback. So the names follow from the semantics of the filter, not from an accident of one environment.

**Evidence.** On a freshly provisioned server (`provisioning/setup.ts --db pdpgsql`, matching this job's `setup_services`) plus a client node labelled `environment=dev`:

- three consecutive loads of the filtered dashboard: `["DB Instances","Min DB Uptime"]` every time, `scrolled=2 attached=2 visible=2`
- the new assertion, reading the list from the POM: passes
- negative case, an environment value matching no node: fails with `Metrics without data are: …`, naming all 58 panels

That last one is the real gain over the old budget - the same condition previously produced a bare number.

**Two things I want to be straight about.**

1. This is not a full green run of the test. `cliHelper.execute` runs on the test-runner host, so `sudo pmm-admin config` cannot execute on my Windows machine; I drove the navigation and assertion directly and left that first line unexercised. The node I measured is also a pmm-client *container* node rather than CI's generic runner node. The reasoning above is node-type independent (it is about which exporters the labelled node runs), but CI is the first place the generic-node shape actually gets exercised.

2. This improves diagnostics, it does not close the filter gap from the earlier thread. If `var-environment` were silently ignored the dashboard renders unfiltered, zero panels are empty, and the allow-list passes just as the count budget did. It catches the label failing to reach the metrics, and now names what is missing when it does.

One caveat for whoever reads `noDataMetrics` next: those two panels are only empty when the selected environment contains no database - unfiltered, as in `homeDashboards.test.ts`, they do have data. Nothing else reads the field today, and an over-generous allow-list can only weaken a check rather than falsely fail one, but it would be the wrong list to reuse for an unfiltered assertion.

travagliad (approved):

